PR #8, opened 2022-04-04
Allow files to be downloaded from private wikis

### LINKED ISSUE #7: Support private repository wikis
Right now, if a repository's wiki is private, this action will not work, since private wikis cannot have their content accessed using raw.githubusercontent.com.

Doggie52:
Closes #7 .

### DIFF
--- a/coverage_comment/wiki.py
+++ b/coverage_comment/wiki.py
@@ -1,3 +1,4 @@
+import contextlib
 import pathlib
 import tempfile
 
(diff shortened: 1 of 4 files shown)


Doggie52:
Testing this would require building a mocker for `git clone` which stores the file you want to upload - not entirely sure how to do this so would appreciate guidance.

coverage_comment/wiki.py
```diff
@@ -67,6 +69,27 @@ def get_file_contents(
...
+                dir = pathlib.Path(dir_path)
+                git.cwd = dir
+
+                git.clone(
+                    f"https://x-access-token:{github_token}@github.com/{repository}.wiki.git",
+                    ".",
+                )
+
+                try:
+                    return (dir / filename).read_text()
```
ewjoachim, line 89: This is mostly copy-pasted from above. D you think we could refactor that ? For example as a `@contextlib.contextmanager` that would clone the repo and `yield dir` and you'd do `with cloned_wiki(...) as dir:` ?

Also, I believe there is a good coverage currently, so you may want to add some additional tests ?

coverage_comment/wiki.py
```diff
@@ -67,6 +69,27 @@ def get_file_contents(
...
+                except FileNotFoundError:
+                    pass
```
ewjoachim, line 91: Maybe worth adding another log here ?

coverage_comment/wiki.py
Doggie52: I have added tests - I can't trigger the tests to run and I haven't set up local testing so can't see whether it will pass.

Doggie52:
Tests pass, coverage is 100%. The remaining failure I believe is because I am the one running the tests and not you, I think.

ewjoachim:
Thank you, and congratulations :)

Issue #11, opened 2022-05-11
Always getting "100% of new lines are covered"


acwl-aiq:
Thanks for writing this interesting GH action script. I was experimenting with it in a sandbox repo. The line and branch coverage appears to be working as intended but the new line coverage is always showing 100%. 

I probably did something wrong in the configuration that I wasn't able to immediately notice. I wonder if you can lend an eye on this?

This is a pull request where I deliberately introduced code that's not covered by tests. https://github.com/acwl-aiq/lambda_lock_poc/pull/7

Additionally, is there a config I can use to make decreases in line/branch coverage to show red instead of blue? Thanks!

ewjoachim:
Hm, strange. I don't have an idea right now, I'll need to give it a look later. No promises, but I'll see what I can do.

acwl-aiq:
That's fine. This is not urgent

kieferro:
I have copied the repo and looked at the logs. It seems that the coverage module, when using 
```
source =
        lib
        lambda_lock_poc
        construct
```
in the `.coveragerc` file only writes relative paths regarding the sources into the XML file and the `diff-cover` module then can't handle this and doesn't find the files. With only new lines without coverage information the module outputs 100% coverage. (It might be useful to write in the PR-comment in such a case that no lines with coverage information were created instead of 100%. I could implement that, if you want @ewjoachim ).

I'm not quite sure if there's a bug in there somewhere or if this is all expected behavior, but one workaround I've found is to take the source information out of the .coveragerc file and instead write into the workflow file which folders to test. So instead of this line:
`coverage run --rcfile=.coveragerc -m pytest`
this one:
`coverage run --rcfile=.coveragerc -m pytest ./lib/ ./lambda_lock_poc/ ./tests/`

I tested this with a copy and it worked for me so far. Would this be a satisfactory solution?

Regarding the red arrow, there is no setting. However, I could implement it so that you can specify which emojis should be displayed when the coverage increases/decreases. (Of course, only if @ewjoachim thinks it makes sense to implement this feature).

ewjoachim:
Starting with the easy answer:

> Regarding the red arrow, there is no setting. However, I could implement it so that you can specify which emojis should be displayed when the coverage increases/decreases. (Of course, only if @ewjoachim thinks it makes sense to implement this feature).

[...]

PR #21, opened 2022-06-27
Add possibility to extend blocks of the comment template

kieferro:
This adds the functionality for the `COMMENT_TEMPLATE` parameter as discussed in #11. Also, this parameter has been added to the README.


tests/integration/test_main.py
```diff
@@ -370,3 +370,46 @@ def test_action__workflow_run__post_comment(
     assert result == 0
     assert get_logs("INFO", "Comment file found in artifact, posting to PR")
     assert get_logs("INFO", "Comment posted in PR")
+
+
+def test_action__pull_request__comment_template(
```
ewjoachim, line 375: When a fixture is only relevant for one specific test, I prefer putting the fixture next to the test.
Also, I'm not sure why we're using a fixture here ? Wouldn't it be clearer to define the `COMMENT_TEMPLATE` in the test directly ?

Finally... This is a high level intergation test. Wouldn't it be easier to implement this test as a lower-level unit test (in https://github.com/ewjoachim/python-coverage-comment-action/blob/v2/tests/unit/test_template.py )

kieferro: I originally thought it would be clearer to implement all configurations as a fixture. But I absolutely understand that it's not worth it for a one-time use. I have fixed that.

The way I see it, testing if the `get_markdown_comment` function can handle templates is already implemented in `test_template`. But this test should test the `clean_comment_template` function, which was also correctly shown as not covered before. But I didn't see any other way than to implement that high-level, because that is only called when `COMMENT_TEMPLATE` is set in the configuration. How would I implement this as a low-level unittest?

kieferro: I could, of course, write a test that simply calls `Config.from_environ` once with and once without `COMMENT_TEMPLATE` and then checks to see if the correct value was stored. Is that what you mean?

coverage_comment/settings.py
```diff
@@ -27,6 +29,7 @@ class Config:
     GITHUB_REF: str
     GITHUB_EVENT_NAME: str
     GITHUB_PR_RUN_ID: int | None
+    COMMENT_TEMPLATE: str = template.read_template_file()
```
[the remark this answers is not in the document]
kieferro: I can change that, but that would break all tests that don't specify `COMMENT_TEMPLATE`, because then the clean method is never called. (When I run it locally in a Docker container, 4 tests fail because of this change).
Should I change it anyway and rewrite the tests to read the template from the file and pass it to the configuration?

tests/integration/test_main.py
kieferro: Done

ewjoachim:
I've started working on a PR-on-your-PR to address different points, I'll try to post it in the close future. Being sick makes it complated to evaluate how much clear-mind time I'll have on my hands in the next days (and how much I'll dedicate to coding 😅 )

ewjoachim:
@kieferro aaaaand done. I've reworked the PR and added my suggestions (working with blocks and template inheritance, documentation, etc). Also, I included a fix for the issue that currently makes the CI red.

PR #23, opened 2022-07-03
Fix examples in readme

kieferro:
I've been messing around a bit with `COMMENT_TEMPLATE` in my private repository and noticed that the examples in the `README` don't all work or cause unwanted effects (visible marker).

Regarding the quotes in the last commit: when I tried to use the example without quotes, Github did not interpret the input as a string, but tried to process it somehow differently. While this is not a problem with some other examples, it might make sense to make all examples with quotes and add a note to that effect, but that's their decision.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ In this second example, we replace the whole comment by something much shorter w
 coverage (percentage) of the whole project from the PR build:
 
 ```jinja2
-Coverage: {{ coverage.info.percent_covered | pct }}
+Coverage: {{ coverage.info.percent_covered | pct }}{{ marker }}
 ```
 
 # Other topics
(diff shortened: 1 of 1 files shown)


ewjoachim (approved):

ewjoachim:
Thank you very much :)

Hey, I'm noticing you're quite active on the project, and I'm not that much. Would you be interested with an official "position" in the project ? Not that it's something with lots of activity (as you've seen), but you seem interested, so I'm asking :)

kieferro:
Sure, glad to help :)

First of all, I am very honored. Thank you for the offer. What would that mean explicitly?
I am happy to take some responsibility for the project and help especially with opened issues or questions.

ewjoachim:
> What would that mean explicitly?

I can give you maintainer rights to the repo, or commit permissions if you want, list you on the readme as a maintainer with me, set you as a codeowner, etc. You tell me :) You're in the team :)

kieferro:
> You're in the team :)

Thank you :)

Commit permission would be great to be able to directly fix small bugs (like the one in this PR). For big changes I would still open PRs to get your approval.
You are also welcome to write me in the README and add me as codeowner.

PR #25, opened 2022-07-10
Fix pull request comment for differently named forks

kieferro:
Currently, a pull request is not found if the repo it comes from is named differently than the main repository. This is because when specifying `head` with `username:branch`, the Github API only looks for repos with the same name. To fix this, you need to specify `username/reponame:branch`.
This PR fixes the problem and changes the tests to match the change.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
   - id: check-added-large-files
 
 - repo: https://github.com/psf/black
-  rev: "21.12b0"
+  rev: "22.6.0"
   hooks:
   - id: black
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
   - id: check-added-large-files
 
 - repo: https://github.com/psf/black
-  rev: "22.3.0"
+  rev: "21.12b0"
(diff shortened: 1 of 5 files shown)


ewjoachim (approved):
Seems awesome !

.pre-commit-config.yaml
```diff
@@ -13,7 +13,7 @@ repos:
   - id: check-added-large-files
 
 - repo: https://github.com/psf/black
-  rev: "22.3.0"
+  rev: "21.12b0"
```
ewjoachim, line 16: I think this line was auto-added by pre-commit to sync with poetry, that's my fault. Cal you run `poetry update` and commit the changes in lockfile & `pre-commit-config.yaml`` ? That will fix the CI.

kieferro: I did that, but `poetry update` changed the file a lot. I am not quite sure if this is how it should be. I have committed the changes, but feel free to revert the changes or let me know if you want me to delete the commit.

ewjoachim: Nah, updating the lockfile should never be an issue. If it was, it would mean that our declared versions in pyproject.toml are incorrect.

ewjoachim:
congratulations :tada:

PR #26, opened 2022-07-11
Fix code for internal branches

kieferro:
Sorry, I used the wrong parameter name #25

### DIFF
--- a/coverage_comment/github.py
+++ b/coverage_comment/github.py
@@ -76,7 +76,7 @@ def get_pr_number_from_workflow_run(
     repo_path = github.repos(repository)
     run = repo_path.actions.runs(run_id).get()
     branch = run.head_branch
-    repo_name = run.head_repository.name
+    repo_name = run.head_repository.full_name
     full_branch = f"{repo_name}:{branch}"
     get_prs = functools.partial(
         repo_path.pulls.get,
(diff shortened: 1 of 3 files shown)


ewjoachim:
... I'll merge this, but I reckon the issue is that even with 100% unit coverage, we need real end to end tests. Maybe at some point we'll need automated tests, but one way or another, it would be nice if we could manually test this before merging in the meantime.

That being said, I trust you that this will actually solve the issue :)

ewjoachim (approved):

kieferro:
Yes, sorry again. I really didn't mean to break the code. But somewhere between testing locally with the API and changing the code, I must have copied the wrong name.
I just tested again in my own project where I had noticed the error too and everything is running correctly again now. So the error should be fixed.
Real end-to-end testing would certainly be very helpful, but I also understand that it's difficult to set up. I will test more extensively locally in the future.

ewjoachim:
1. Don't beat yourself up, it's not your fault. If the process (the automated CI, my review, etc) left enough room for a bug to arise, then it's the process that's faulty. That was an honest mistake :)

2. I'm not sure that it would be that hard to setup something automated, especially using the `gh` cli. That said, it might be easier if it runs post-merge, at least for a start.

PR #36, opened 2022-07-29
Last update to v2, new developments will be on v3

ewjoachim:
@kieferro Please feel free to tell me if I forgot something important (or not important) :)

I think the plan is:
- Fork the v3 branch ([done already](https://github.com/ewjoachim/python-coverage-comment-action/tree/v3))
- On v3, have Docker don't push the `latest` docker tag.
- Merge this
- Then switch v3 as the default branch
- Then merge #33 
- Then work on the end-to-end tests on v3
- Then maybe release v3.0.0 and re-enable latest

PR #40, opened 2022-07-31
Pull a "static" image, build the latest code on top

### LINKED ISSUE #38: Improve the link between action version and docker image version
## Build Dockerfile seemlessly

It would be nice if we could read a local variable (DEBUG or BUILD_FROM_DOCKERFILE), (an action parameter that could be set  set as repository variable or hardcoded in the workflow), that would switch from pulling the image to building it from the Dockerfile.

I'm sure we're not the first ones doing that, so maybe we can reuse something.

## Use correct version

I believe that, as of today, using a release tag of the version still ends up using the latest docker version, so it's probably quite useless.

ewjoachim:
Closes #38 

This should have multiple advantages:
- We should still run quite quickly. Hopefully less than 10s
- Now, if the action runs on git ref X, it means the code underneaath *really* is the code that ref X points to is used.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -39,6 +39,7 @@ jobs:
         uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
         with:
           context: .
+          # Bump when changing anything used in Dockerfile
+          tags: v1
           push: true
-          tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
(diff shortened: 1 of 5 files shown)


kieferro:
I apologise if you were waiting for my approval. Everything looks good to me so far. I think we can merge and then proceed according to the plan in #36.

ewjoachim:
> I apologise if you were waiting for my approval

Nah, don't worry, it's just that some weeks I have a lot of time on my hands, and some weeks I don't :)

PR #45, opened 2022-08-10
fix/white-space 

ME-ON1:
for `100%` file coverage ( when there is no violation_lines ), there is a issue with whitespacing, the file title following, are concatenated with the prev violating_lines block.

This PR aims to solve that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,17 +12,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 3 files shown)


ewjoachim:
I'm not 100% sure that your solution is the best one compared to removing the `-` at the end of `{% block coverage_single_file scoped -%}` but I think in the end, it's not very different. Jinja and whitespaces is confusing.

Would you mind writing a test ?

ewjoachim:
I believe there's already a test around this question: https://github.com/ewjoachim/python-coverage-comment-action/blob/v2/tests/unit/test_template.py#L54-L62

Any idea why it didn't show the issue ?

ewjoachim:
So I added the necessary tests, I'll merge as soon as it's green.

As a side note: when contributing from a fork, it help if you name your branch with a name different from the default branch of the original repo. I almost pushed to _this repo's_ v2 instead of _your fork's_ v2.

PR #55, opened 2022-09-04
Make end-to-end tests on PR (running only for admins) use the current commit

ewjoachim:
When GitHub launches the end-to-end tests of the actions in the CI, it will now ensure that the version of of the action that runs in the end-to-end repo is the one from the PR, so we really test the action.


ewjoachim:
Woo it works! When launched on a [commit](https://github.com/ewjoachim/python-coverage-comment-action/commit/6415542e9a6bd7a2c750fa77a4da08860fb50298) that fails the action, the end-to-end tests in the PR [failed](https://github.com/ewjoachim/python-coverage-comment-action/runs/8177107282?check_suite_focus=true) \o/

PR #58, opened 2022-09-06
Change badge to v2

kieferro:


ewjoachim:
Why are we using the v2 badge in the v3 branch ?

kieferro:
> Why are we using the v2 badge in the v3 branch ?

Because the workflow we use also still uses `v2` (https://github.com/ewjoachim/python-coverage-comment-action/blob/v3/.github/workflows/ci.yml#L46). I can also create a PR to update the workflow to `v3`, but I thought you might want us to keep `v2` for this repo, because you haven't updated the workflow with the PR that created `v3`.

PR #61, opened 2022-09-07
Update end-to-end-tests to use PR number instead of branch name

kieferro:
This PR changes the end-to-end tests, to use the PR number instead of the name of the branch as the suffix for the created repo. This has several advantages:
- The name of the repo is not longer than the allowed maximum length.
- The name of the repo clearly indicates which PR it refers to.
- There is no chaos with branches of the same name from different repos.

Do we have to change anything for `push` or anything else @ewjoachim?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
     branches:
       - 'v3'
 
+concurrency:
+  group: ${{ github.event_name }}-${{ github.ref }}
+
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


kieferro:
Concurrency might also be good to ensure that two end-to-end tests never run on the same branch at the same time. Something like this:
```yml
concurrency:
  group: ${{ github.head_ref || github.ref_name }} 
```

ewjoachim (approved):

PR #67, opened 2022-09-14
Introduce workflow_run event for Docker publish

### LINKED ISSUE #64: Don't publish the docker image if the CI fails
we need a way to avoid publishing the docker image if the ci fails. Probably using `on: workflow_run` and restricting on:
- CI workflow being successful
- CI workflow event to be a push on the v3 branch (and not a PR targetting the v3 branch and/or based on a head branch also named v3 but on a fork)

kieferro:
Closes #64 

With this change, the Docker Publish should only be executed if the tests were successful.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -6,10 +6,6 @@
 name: Publish Docker image
 
 on:
-  release:
-    types:
-      - published
-
   workflow_run:
     workflows: [CI]
     types: [completed]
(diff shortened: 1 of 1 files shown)


kieferro:
I have checked again in a test repo that the condition works for both failed tests and pull_request events and so everything should work.
Have I forgotten anything else @ewjoachim?

.github/workflows/docker.yml
```diff
@@ -9,14 +9,18 @@ on:
   release:
     types:
       - published
```
ewjoachim, line 11: I guess we should remove that part too, no ?

ewjoachim (approved):
Good for me !

PR #71, opened 2022-09-26
Various small things

ewjoachim:
- I ended up finding it annoying that the e2e test repo was at the root of this repo. Moving it somewhere more appropriate
- There's an unused requirements.txt file. Removing it.


ewjoachim:
Hm, same error as #70 . No idea why it started failing. Something to do with Poetry 1.2 (I doubt but it's possible).

kieferro (approved):

kieferro:
(pushed empty commit with `--force` to force workflow run with new file #73)

PR #84, opened 2022-10-15
Use environment files instead of `set-output`

### LINKED ISSUE #83: GitHub Actions set-output is deprecated
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

We need to change our way of setting an output value.

```python
if settings.OUTPUT_FILE:
    with open(settings.OUTPUT_FILE, "a") as f:
        f.write("a=b")
```

kieferro:
Closes #83


coverage_comment/github.py
```diff
@@ -153,6 +153,8 @@ def post_comment(
...
 
 
-def set_output(**kwargs: bool) -> None:
-    for key, value in kwargs.items():
-        print(f"::set-output name={key}::{json.dumps(value)}")
+def set_output(config: settings.Config, **kwargs: bool) -> None:
+    if config.GITHUB_OUTPUT:
+        with open(config.GITHUB_OUTPUT, "a") as f:
+            for key, value in kwargs.items():
+                f.write(f"{key}={json.dumps(value)}\n")
```
ewjoachim, line 160: rather than passing the whole config, maybe we could just pass `config.GITHUB_OUTPUT` so that the function would be simpler. (and easier to test)

coverage_comment/main.py
```diff
@@ -162,10 +162,10 @@ def generate_comment(
             filename=config.COMMENT_FILENAME,
             content=comment,
         )
-        github.set_output(COMMENT_FILE_WRITTEN=True)
+        github.set_output(config, COMMENT_FILE_WRITTEN=True)
         log.debug("Comment stored locally on disk")
     else:
-        github.set_output(COMMENT_FILE_WRITTEN=False)
+        github.set_output(config, COMMENT_FILE_WRITTEN=False)
```
ewjoachim, line 168: just a note: I see you pass config by position rather than by name (`config=config`).
On functions that have multiple parameters, I almost always use named arguments.

I guess one thing you might or might not be wondering is whether this might interfere with this function defining `**kwargs`, so just to be extra-clear: `**kwargs` contains all the keyword arguments that were not bound to "normal" arguments, so you can use a keyword arg here.

tests/conftest.py
```diff
@@ -385,3 +386,15 @@ def _(*, exit_code=0, stdout=""):
     git = Git()
     yield git
     assert not git.expected_calls
+
+
+@pytest.fixture
+def output_file():
```
ewjoachim, line 392: please either:
- use the `tmp_path` pytest fixture
- use a fixture that uses it

The idea being that we don't want this to happen in the current repo. Also, you won't have to clean afterwards which make things easier to debug (pytest will leave old folders around for the last 3 invocations before deleting)

tests/conftest.py
```diff
@@ -385,3 +386,15 @@ def _(*, exit_code=0, stdout=""):
...
+    FILENAME = "temp_output.txt"
+
+    with open(FILENAME, "w") as _:
+        pass
+
+    yield FILENAME
+
+    os.remove(FILENAME)
```
ewjoachim, line 400:
```suggestion
    file = pathlib.Path("temp_output.txt")

    file.touch()

    yield file

    file.unlink()  # (though you don't need to do it if you use tmp_path)
```

tests/integration/test_github.py
```diff
@@ -288,7 +288,8 @@ def test_post_comment__update_error(gh, session):
...
 
-def test_set_output(capsys):
-    github.set_output(foo=True)
-    captured = capsys.readouterr()
-    assert captured.out.strip() == "::set-output name=foo::true"
+def test_set_output(push_config, output_file):
+    github.set_output(config=push_config(GITHUB_OUTPUT=output_file), foo=True)
+
+    with open(output_file) as file:
+        assert file.read() == "foo=true\n"
```
ewjoachim, line 295: Assuming you took my earlier suggestion, output_file will be a `pathlib.Path` object

```suggestion
assert output_file.read_text() == "foo=true\n"
```

coverage_comment/settings.py
```diff
@@ -29,6 +29,7 @@ class Config:
     GITHUB_TOKEN: str = dataclasses.field(repr=False)
     GITHUB_REPOSITORY: str
     GITHUB_REF: str
+    GITHUB_OUTPUT: str
```
ewjoachim, line 32:
```suggestion
    GITHUB_OUTPUT: pathlib.Path | None = None
```

See `COMMENT_FILENAME` below and `clean_comment_filename` below for a way to handle this as a `Path()` instead of a string. This will make the rest of the code easier.

I'm also suggesting adding a default value to None (so you'll need to move this line a few lines down, because you can't have attribute with default values above attributes without default values)

tests/integration/test_main.py
```diff
@@ -179,12 +181,14 @@ def checker(payload):
     assert not pathlib.Path("python-coverage-comment-action.txt").exists()
     assert "The coverage rate went from `30%` to `86%` :arrow_up:" in comment
 
-    expected_stdout = "::set-output name=COMMENT_FILE_WRITTEN::false"
-    assert capsys.readouterr().out.strip() == expected_stdout
+    expected_environment = "COMMENT_FILE_WRITTEN=false\n"
```
ewjoachim, line 184:
```suggestion
    expected_output = "COMMENT_FILE_WRITTEN=false\n"
```

tests/integration/test_main.py
```diff
@@ -179,12 +181,14 @@ def checker(payload):
...
+
+    with open(output_file) as file:
+        assert file.read() == expected_environment
```
ewjoachim, line 187:
```suggestion
    assert output_file.read_text() == expected_environment
```

tests/integration/test_main.py
```diff
@@ -203,8 +207,10 @@ def test_action__pull_request__force_store_comment(
 
     assert pathlib.Path("python-coverage-comment-action.txt").exists()
 
-    expected_stdout = "::set-output name=COMMENT_FILE_WRITTEN::true"
-    assert capsys.readouterr().out.strip() == expected_stdout
+    expected_environment = "COMMENT_FILE_WRITTEN=true\n"
+
+    with open(output_file) as file:
+        assert file.read() == expected_environment
```
ewjoachim, line 213: (same as above)

tests/unit/test_settings.py
```diff
@@ -28,6 +28,7 @@ def test_config__from_environ__ok():
             "GITHUB_TOKEN": "foo",
             "GITHUB_REPOSITORY": "owner/repo",
             "GITHUB_REF": "master",
+            "GITHUB_OUTPUT": "",
```
ewjoachim, line 31: I think the idea is to put as many non-default values here as possible and check that were properly transcribed below, so I'd go for a non-empty string.

tests/unit/test_settings.py
```diff
@@ -45,6 +46,7 @@ def test_config__from_environ__ok():
         GITHUB_TOKEN="foo",
         GITHUB_REPOSITORY="owner/repo",
         GITHUB_REF="master",
+        GITHUB_OUTPUT="",
```
ewjoachim, line 49: (so this will be a pathlib.Path, like `COMMENT_FILENAME`)

ewjoachim (commented):
I see you have still to convert to the wonderful life of using `pathlib.Path` instead of `os.path`.

Capture d’écran 2022-10-16 à 10 55 49

kieferro:
Thanks for the review. If I haven't forgotten anything, I should have adopted all the suggested changes.

ewjoachim (approved):
Yay, that seems awesome! Smells like modern Python too :D

PR #95, opened 2022-11-01
Add more detailed text to the PR comment

### LINKED ISSUE #51: Change behaviour when missing coverage data for the default branch
It might be helpful to write in the PR-comment that there is no comparative data in PR mode if no coverage data for the default branch was found, instead of crashing the action. This would not lead to so much confusion when the action is set up by a PR (#15, #48).
This is nothing urgent, but something I wanted to put here so that I don't forget it.

kieferro:
Closes #51 
May be clearer in cases like #11

This PR does two things:
* If no comparison data was found, it generates an extra message in the PR comment (#51)
* I found the message "100% of new lines are covered" a bit misleading when there is no coverage information about any of the new lines (since they belong to README, config files or the like, for example). Therefore, this is replaced by another message in this case.

### DIFF
--- a/coverage_comment/default.md.j2
+++ b/coverage_comment/default.md.j2
@@ -14,6 +14,11 @@ The coverage rate went from `{{ previous_coverage_rate | pct }}` to `{{ coverage
 {%- endif %}
 {%- endblock emoji_coverage -%}
 {%- else -%}
+{% block no_comparison_info -%}
+> **Note**
+> No coverage data of the default branch was found for comparison. A possible reason for this is that the coverage action has not yet run after a push event and the data is therefore not yet initialized.
+{%- endblock no_comparison_info %}
+
 {% block coverage_value_wording -%}
 The coverage rate is `{{ coverage.info.percent_covered | pct }}`.
 {%- endblock coverage_value_wording %}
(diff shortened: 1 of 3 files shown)


coverage_comment/default.md.j2
```diff
@@ -28,7 +33,11 @@ The branch rate is `{{ (coverage.info.covered_branches / coverage.info.num_branc
 {% endblock branch_coverage -%}
 
 {% block diff_coverage_wording -%}
+{%- if diff_coverage.total_num_lines > 0 -%}
 `{{ diff_coverage.total_percent_covered | pct }}` of new lines are covered.
+{%- else -%}
+_None of the new lines are part of the tested code. Therefore, there is no coverage data about them._
+{%- endif %}
```
ewjoachim, line 40: The `{% block diff_coverage_wording -%}` should be inside the block, I think, not outside. And we probably need to introduce another block, like `{% block diff_coverage_empty_wording -%}`

ewjoachim (approved):
Apart from my small comment, this seams great! thanks :)

kieferro:
Very good, thanks for the review. I have made the change.

ewjoachim:
:heart:

PR #96, opened 2022-11-03
[New feature] Optional annotations for not covered lines on pull requests

kieferro:
This PR extends the action with a setting that makes it possible that all not covered lines can be [annotated](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message) in PR mode.

### DIFF
--- a/coverage_comment/annotations.py
+++ b/coverage_comment/annotations.py
@@ -1,0 +1,18 @@
+from coverage_comment import coverage as coverage_module
+from coverage_comment import github
+
+MISSING_LINES_GROUP_TITLE = "Annotations of lines with missing coverage"
+
+
+def create_pr_annotations(annotation_type: str, coverage: coverage_module.Coverage):
+    github.send_workflow_command(
+        command="group", command_value=MISSING_LINES_GROUP_TITLE
+    )
+
+    for filename, file_coverage in coverage.files.items():
+        for missing_line in file_coverage.missing_lines:
+            github.create_missing_coverage_annotation(
+                annotation_type=annotation_type, file=filename, line=missing_line
+            )
+
+    github.send_workflow_command(command="endgroup", command_value="")
--- a/coverage_comment/github.py
+++ b/coverage_comment/github.py
@@ -8,6 +8,7 @@ import zipfile
 from coverage_comment import github_client, log
 
 GITHUB_ACTIONS_LOGIN = "github-actions[bot]"
+MISSING_COVERAGE_MESSAGE = "This line has no coverage"
 
 
 class CannotDeterminePR(Exception):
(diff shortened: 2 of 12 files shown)


tests/end_to_end/repo/.github/workflows/ci.yml
```diff
@@ -31,6 +31,7 @@ jobs:
         uses: py-cov-action/python-coverage-comment-action@__ACTION_REF__
         with:
           GITHUB_TOKEN: ${{ github.token }}
+          ANNOTATE_MISSING_LINES: true
```
ewjoachim, line 34: nice ! do you think we should check that they were created properly or is it just intended for humans to have a look ?

ewjoachim (commented):
That seems awesome :) A few comments

coverage_comment/github.py
```diff
@@ -158,3 +159,21 @@ def set_output(github_output: pathlib.Path, **kwargs: bool) -> None:
...
+
+    if len(values_listed) > 0:
+        print(f"::{command} {','.join(values_listed)}::{command_value}")
+    else:
+        print(f"::{command}::{command_value}")
+
+
+def create_missing_coverage_annotation(file: str, line: int):
+    send_workflow_command(
+        command="warning",
```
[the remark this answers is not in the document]
kieferro: Good idea! I implemented that

coverage_comment/annotations.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def create_pr_annotations(coverage: coverage_module.Coverage):
+    github.send_workflow_command(
+        command="group", command_value=MISSING_LINES_GROUP_TITLE
+    )
+
+    for filename, file_coverage in coverage.files.items():
+        for missing_line in file_coverage.missing_lines:
+            github.create_missing_coverage_annotation(filename, missing_line)
```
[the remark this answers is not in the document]
kieferro: right, i must have missed that here

tests/end_to_end/repo/.github/workflows/ci.yml
kieferro: It was mainly meant for me to actually see the feature in action in a repo, so I know if everything works the way I envision it.
I don't think we necessarily need to verify that the annotations work (I don't know how easy that would be either), but if you already have a concrete idea how to do that, we could include it.

coverage_comment/settings.py
```diff
@@ -40,6 +40,7 @@ class Config:
     MINIMUM_ORANGE: float = 70.0
     MERGE_COVERAGE_FILES: bool = False
     ANNOTATE_MISSING_LINES: bool = False
+    ANNOTATION_TYPE: str = "warning"
```
ewjoachim, line 43: I wonder if it wouldn't make more sense to check that we're within the correct values in `clean_annotation_type` rather than later ?
(if someone installs pcca on their repo and configure it incorrectly, they'd probably prefer it to crash immediately rather than at the first uncovered line)

kieferro: Yes, you're right. That makes more sense

kieferro:
> Do we want to install it on the repo itself ?

Sure, let's do that

ewjoachim:
Thank you a lot, that's awesome !

PR #105, opened 2022-11-28
Use `decimal.Decimal` instead of `float`

### LINKED ISSUE #37: Use decimals/fractions everywhere
Floating-point types make knowing whether the coverage increased, decreased or stayed the same quite complicated (mainly: when it stays the same, floating point errors can make us think it changed).

There's [a bit of work](https://github.com/ewjoachim/python-coverage-comment-action/pull/33/commits/08619a78de77bb96cd4609de2650ec3081fb9c4b#diff-61bf637d10e8406acddf352767e17c4c775522b38e64c5f0e065750bcc1d8014R27-R35) in #33 but I guess we'd gain from using consistently fractions or decimals everywhere applicable.

kieferro:
Closes #37

This PR replaces all previous usages of `float` with `decimal.Decimal` to prevent floating-point errors.
I've tried to change all the things adequately, but it's still quite possible that I missed a spot that actually needs a change.
This also allows us to present coverage data more accurately:

### DIFF
--- a/coverage_comment/coverage.py
+++ b/coverage_comment/coverage.py
@@ -1,5 +1,6 @@
 import dataclasses
 import datetime
+import decimal
 import json
 import pathlib
 import tempfile
@@ -19,7 +20,7 @@ class CoverageMetadata:
 class CoverageInfo:
     covered_lines: int
     num_statements: int
-    percent_covered: float
+    percent_covered: decimal.Decimal
     missing_lines: int
     excluded_lines: int
     num_branches: int | None
@@ -47,7 +48,7 @@ class Coverage:
 @dataclasses.dataclass
 class FileDiffCoverage:
     path: str
-    percent_covered: float
+    percent_covered: decimal.Decimal
     violation_lines: list[int]
 
 
@@ -55,9 +56,15 @@ class FileDiffCoverage:
 class DiffCoverage:
     total_num_lines: int
     total_num_violations: int
-    total_percent_covered: float
+    total_percent_covered: decimal.Decimal
     num_changed_lines: int
     files: dict[pathlib.Path, FileDiffCoverage]
+
+
+def compute_coverage(num_covered: int, num_total: int) -> decimal.Decimal:
+    if num_total == 0:
+        return decimal.Decimal("1")
+    return decimal.Decimal(num_covered) / decimal.Decimal(num_total)
 
 
 def get_coverage_info(merge: bool) -> Coverage:
@@ -139,7 +146,12 @@ def extract_info(data) -> Coverage:
                 info=CoverageInfo(
                     covered_lines=file_data["summary"]["covered_lines"],
                     num_statements=file_data["summary"]["num_statements"],
-                    percent_covered=file_data["summary"]["percent_covered"] / 100,
+                    percent_covered=compute_coverage(
+                        file_data["summary"]["covered_lines"]
+                        + file_data["summary"].get("covered_branches", 0),
+                        file_data["summary"]["num_statements"]
+                        + file_data["summary"].get("num_branches", 0),
+                    ),
                     missing_lines=file_data["summary"]["missing_lines"],
                     excluded_lines=file_data["summary"]["excluded_lines"],
                     num_branches=file_data["summary"].get("num_branches"),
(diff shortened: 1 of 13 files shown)


coverage_comment/template.py
```diff
@@ -69,5 +70,11 @@ def read_template_file() -> str:
     return resources.read_text("coverage_comment", "default.md.j2")
 
 
-def pct(val):
-    return f"{val:.0%}"
+def pct(val: decimal.Decimal | float) -> str:
+    if type(val) == decimal.Decimal:
+        val *= decimal.Decimal("100")
+        return (
+            f"{val.quantize(decimal.Decimal('1'), rounding=decimal.ROUND_HALF_DOWN)}%"
```
ewjoachim, line 77: I notice that it was `ROUND_DOWN` and it's now `ROUND_HALF_DOWN`.

I think the only thing I'd like to make sure is that we never display 100% unless we actually have all the lines covered. Is this the case here ? (99.9999999999% should round to 99%, not 100%)

ewjoachim (approved):
Nice :) Just a single comment. Congratz !

coverage_comment/template.py
kieferro: Thank you for the review!

I'm not quite sure that I understand this comment. Before, we used `ROUND_DOWN` for creating the badges, but this code here is for the template. The previous implementation
```python
def pct(val):
    return f"{val:.0%}"
```

also rounds up or down, depending on how large the decimal place is. I guess this was not intended then?
Whereby this is a bit strange, because also some of the tests seem to assume that the values in the templates aren't always rounded down. This line [here](https://github.com/py-cov-action/python-coverage-comment-action/blob/3f19483f8228c96f9ed1964ed93359b94e87b5fe/tests/integration/test_main.py#L128) should actually assume 85% coverage when rounding down, if I haven't made a bad miscalculation. So we actually also want to round down in the template and this is just an error?

ewjoachim: Ah, well, maybe it was wrong before...?

What's your gut feeling on this ?

kieferro: > Ah, well, maybe it was wrong before...?

In fact, I'm pretty sure it must have been wrong before that. When we lost the 100% coverage for this repo, the comment still showed 100% [as new coverage](https://github.com/py-cov-action/python-coverage-comment-action/pull/84#issuecomment-1279857308).


> What's your gut feeling on this ?

I definitely agree that it is not good to display 100% unless there really is 100%. For the badges, the previous solution with rounding down definitely works. But I'm not sure if that's the best solution for the comment as well. Now that we actually have the percentages exactly and no longer use float, would there be anything against displaying all percentages in the comment with two (I think two would be the best middle ground between inaccurate and unnecessarily accurate) decimal places and rounded down? Then one would have an exact indication, also for the comparison, and it would be nevertheless only 100% indicated, if also 100% are there. What are your thoughts on this?

ewjoachim: 4 significant digits + rounding down in the comments, that seems good to me

coverage_comment/coverage.py
```diff
@@ -139,7 +140,10 @@ def extract_info(data) -> Coverage:
                 info=CoverageInfo(
                     covered_lines=file_data["summary"]["covered_lines"],
                     num_statements=file_data["summary"]["num_statements"],
-                    percent_covered=file_data["summary"]["percent_covered"] / 100,
+                    percent_covered=decimal.Decimal(
+                        str(file_data["summary"]["percent_covered"])
```
ewjoachim, line 144: If we're putting precision in this thing, shouldn't we recompute percent_covered as `covered_lines/num_statements` ? I'm guessing we may inherit rounding errors otherwise.

kieferro: Yes, that is a very good point :+1:.
Are you sure that this is the calculation formula that is also used internally? Don't we still have to include the branches somehow? Because otherwise one of our [test cases](https://github.com/py-cov-action/python-coverage-comment-action/blob/v3/tests/conftest.py#L88) seems to be wrong? 
I have already tried to find documentation for this, but I was not really successful

ewjoachim: `num_statements+num_statements` ?

kieferro: 😄 Copied the wrong bit of code: `(covered_lines+covered_branches)/(num_statements+num_branches)`

ewjoachim: :+1:

kieferro:
@ewjoachim Other than that, do you have any comments or can I squash commit 2 and 4 and then merge?

ewjoachim (approved):
Perfect !

Thank you so much :)

kieferro:
> Thank you so much :)

No problem :) Thank you for reviewing and making sure everything works. It is very reassuring to know that four eyes have looked over the code and not just two.

PR #131, opened 2023-01-31
Add possibility for custom suffix for e2e test repos

kieferro:
This PR makes it possible to add a suffix to the e2e test repo with the `repo_suffix` mark. This way the private and the public repo can be named differently.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,3 +29,5 @@ testpaths =
 
 filterwarnings =
     error
+markers=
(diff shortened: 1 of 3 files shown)


ewjoachim (approved):
I find it awesome how using pytest mark here allowed communicating FROM the test TO the fixture in a simple and readable way, and I love how simple it was to edit the fixture to read the mark and extract args.

PR #135, opened 2023-02-05
Improve Logging: use GHA workflow commands

ewjoachim:
Multiple things appear in this PR, please feel free to request if you're more comfortable if I split it into multiple PRs

- First, I split the method for generating workflow command strings so that we decouple generating the string and printing it. 
- Rename on_coverage_branch > checked_out_branch: kind of unrelated change. I had originally more on there but I decided not to keep it. This one is an isolated change and it annoys me every time I see this function name
- Add a GitHub formatter that formats logs the GitHub way: that's the main thing of the PR: making it so that `logger.error`displays a proper error in the GHA output.
- Deprecate VERBOSE setting: The correct way to add logging to a GitHub Action should be to do it the GitHub way (I've added `ACTIONS_STEP_DEBUG=true` in this repo btw)
- If exception bubbles up to the top, log it properly: While we're at it, if the action fails with an exception, at least make it a pretty one.
Capture d’écran 2023-02-05 à 18 16 10

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,6 @@ jobs:
         with:
           GITHUB_TOKEN: ${{ github.token }}
           ANNOTATE_MISSING_LINES: true
-          VERBOSE: "true"
 
       - name: Store Pull Request comment to be posted
         uses: actions/upload-artifact@v3
--- a/README.md
+++ b/README.md
@@ -264,9 +264,6 @@ jobs:
     # https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-notice-message
     ANNOTATION_TYPE: warning
 
-    # If true, produces more output. Useful for debugging.
-    VERBOSE: false
-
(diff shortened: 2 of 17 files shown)


kieferro (approved):
A very nice idea and a nice reuse of the workflow annotation code. I have absolutely nothing to complain about 😀 

We will lose our 100% coverage, but testing exception handling in the main function is of course difficult and not really necessary. If we still want to execute the _log.info_ code in the _clean_verbose_ function when testing, we could set the [last use of Verbose](https://github.com/py-cov-action/python-coverage-comment-action/blob/improve-logging/tests/unit/test_settings.py#L44) to True. Otherwise, though, we could actually delete that.

ewjoachim:
I'm going to try and see how we can keep our 100% We've come so far...

kieferro:
I took care of it and added two test cases. Now our 100% coverage are safe 😃 . Should we merge then?

ewjoachim:
Aww you're too fast :D

Issue #138, opened 2023-02-06
error: failed to push some refs to... Updates were rejected because the tip of your current branch is behind its remote counterpart


ewjoachim:
https://github.com/py-cov-action/python-coverage-comment-action/actions/runs/4100505236/jobs/7071400413#step:9:57

This is apparently because of #124 but I haven't understood yet why it happens.

EDIT: OK
- we generate the htmlcov folder on the main branch, but we want to commit it on the coverage branch. I hadn't realized that, so I did nothing initially and it worked, because modified files are moved when we checkout
- but as soon as we're working on a directory where we already have committed htmlcov files, it fails because we can't checkout anymore
- we need to save the whole htmlcov dir somewhere, then checkout, then delete the previously committed htmlcov, copy the new one back, git add . and commit

What would be the best way of "saving" the htmlcov dir:
- git stashing it ?
- saving it to /tmp ?
- saving each individual file in memory, in a list of objects ?
- saving a single zip in memory and then expanding it back ?
Meanwhile, I've reverted the HTML report so it doesn't fail everyone's CI

PR #139, opened 2023-02-07
Revert "Revert HTML report"

ewjoachim:
We're back for more. This PR re-introduces HTML coverage:
- re-revert the code
- and introduces a new strategy for moving the code to the right branch:
	- generating the html report outside of the repo directory
	- then switch to the coverage branch (in a more reliable way)
	- then delete the existing htmlcov dir
	- and finally move the new htmlcov dir in place, git add and commit

I'm going to try and test this better


kieferro (changes requested):
Everything looks good in itself. Good work

I thought it would still be helpful if the e2e tests also reflected the problem we had before. Therefore I created a PR that complements this PR (#140). In this case, the first PR is simply merged for the e2e tests. This creates a second commit on the default branch. However, it seems that the problem (#138) still exists (https://github.com/mihcaojwe/python-coverage-comment-action-end-to-end-140-public/actions/runs/4137302843/jobs/7152298129)

coverage_comment/storage.py
```diff
@@ -54,52 +35,50 @@ def checked_out_branch(git: subprocess.Git, branch: str):
...
 
-    branch_exists = True
+    log.debug("Resetting all changes")
+    # Goodbye `.coverage` file.
+    git.reset("--hard")
+
     try:
-        git.checkout(branch)
-        log.debug(f"Branch {branch} exist.")
+        git.rev_parse("--verify", branch)
```
kieferro, line 43: The problem seems to be that the action does not recognize the remote branch and therefore creates its own. This is because rev-parse does not seem to work for a branch that does not also exist locally. We need to either checkout again to find out if the branch exists remotely or find another method.

ewjoachim: I'm going to test, but I think verifying `origin/{branch}` instead of `branch` might suffice.

ewjoachim:
I think we're good now.

kieferro (approved):
Yes, that looks very good now :+1:

PR #144, opened 2023-02-22
Add comment regarding permissions to exception message

kieferro:
Due to the fact that the default settings regarding permissions only differ in newly created repos, the workflow may not work in forks even though it works in the original repo because the permissions are set differently in the fork. In this case, the current error message is relatively unhelpful. This PR is intended to improve that.

### DIFF
--- a/coverage_comment/main.py
+++ b/coverage_comment/main.py
@@ -48,7 +48,7 @@ def main():
 
     except Exception:
         log.exception(
-            "Critical error. Please look for open issues or open one in https://github.com/py-cov-action/python-coverage-comment-action/"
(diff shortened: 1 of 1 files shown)


ewjoachim:
Ok, but didn't the changes we made in the README supposedly add the necessary permissions to the action ? Also, I wonder if there may be a better way (e.g. when failing, actually introspect the permissions and of it's a permission, say that ?

We can merge in the meantime :)

ewjoachim (approved):

kieferro:
> Ok, but didn't the changes we made in the README supposedly add the necessary permissions to the action ?

Yes, they do when you setup the action now. The situation this refers to is that there is a repo that still gives read and write permissions by default because it was created before Github changed the default value for it. That is why the permissions in the config files are not changed there, because they are not necessary. But, if you then fork that, then the permissions default to read only and the action runs fine in the main repo but crashes on the fork.

> when failing, actually introspect the permissions

Yes, that would be the nicer and complete solution, of course.

ewjoachim:
Ah, I get it !

PR #152, opened 2023-04-01
Only invite to e2e-repo when the tests fail

kieferro:


ewjoachim (approved):
Thank you very much :)

PR #155, opened 2023-04-06
Add concurrency to e2e-repo workflow

kieferro:
This PR adds concurrency to the CI workflow in the e2e-test repo. The reason for this is that there are sometimes problems because the action for the initial commit is started multiple times. With concurrency, the workflow should at least no longer run in parallel and thus not cause so many problems.


ewjoachim (commented):
👍

PR #180, opened 2023-05-31
feat: marker as argument

ArzelaAscoIi:
- added the "marker" argument to enable multiple comments for multiple services in a single repository 
- adjusted tests for a custom marker
image

### DIFF
--- a/coverage_comment/template.py
+++ b/coverage_comment/template.py
@@ -7,8 +7,6 @@ from jinja2.sandbox import SandboxedEnvironment
 
 from coverage_comment import coverage as coverage_module
 
-MARKER = """<!-- This comment was produced by python-coverage-comment-action -->"""
-
 
 def uptodate():
     return True
@@ -45,6 +43,7 @@ def get_comment_markdown(
     previous_coverage_rate: decimal.Decimal | None,
     base_template: str,
     custom_template: str | None = None,
+    marker: str = "<!-- This comment was produced by python-coverage-comment-action -->",
 ):
     loader = CommentLoader(base_template=base_template, custom_template=custom_template)
     env = SandboxedEnvironment(loader=loader)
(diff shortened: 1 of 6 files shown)


kieferro (approved):
Thank you for your PR. Looks good to me so far.

An addition to the README would still be nice so that the docs contain the new parameter. I guess this is then also to be used in conjunction with `COVERAGE_DATA_BRANCH` so that coverage data can be collected for several applications in one repo? This could perhaps also be written specifically in the README.

coverage_comment/template.py
```diff
@@ -45,6 +43,7 @@ def get_comment_markdown(
     previous_coverage_rate: decimal.Decimal | None,
     base_template: str,
     custom_template: str | None = None,
+    marker: str = "<!-- This comment was produced by python-coverage-comment-action -->",
```
ewjoachim, line 46: I guess this might be a mandatory parameter, to avoid hardcode the default value in multiple places more than necessary

ewjoachim (commented):
Nice :) Same as the other PR, If you can rework the commits a bit, it would be great but feel free to mention if you'd rather we do it.

ArzelaAscoIi:
Thanks for the review! I will adjust the docs + and rework the commits  next week :)

ewjoachim:
Interestingly, I just realized that I did a branch 4 months ago that I never finished, which did the same thing. Here's the work I did, might inspire you :)

https://github.com/py-cov-action/python-coverage-comment-action/tree/monorepo

ewjoachim:
(Just so you know, I've finally openned the PR https://github.com/py-cov-action/python-coverage-comment-action/pull/213 based on the branch I mentionned above. It does things a bit differently, using a single setting to take care of all monorepo issues)

ArzelaAscoIi:
Ahh for not getting back to it. So let's close this one and follow your approach :) 
Thanks!

PR #201, opened 2023-06-28
Add job summaries for PR and push mode

### LINKED ISSUE #187: Job Summaries
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary

Looks like GitHub now supports posting a markdown that summarizes the run. It just so happen that we do have a markdown ready for this (the comment that we post). Maybe it would make sense that we post our markdown as job summary too ? It probably wouldn't hurt, at least.

kieferro:
Closes #187

The changes in this PR add [job summaries](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary) as a functionality. This means that in PR mode the comment and in push mode the coverage table are also displayed as a job summary.

### DIFF
--- a/coverage_comment/github.py
+++ b/coverage_comment/github.py
@@ -197,10 +197,10 @@ def create_missing_coverage_annotation(annotation_type: str, file: str, line: in
     )
 
 
-def append_to_environment_file(content: str, filepath: pathlib.Path):
+def append_to_file(content: str, filepath: pathlib.Path):
     with filepath.open(mode="a") as file:
         file.write(content)
 
 
 def add_job_summary(content: str, github_step_summary: pathlib.Path):
-    append_to_environment_file(content=content, filepath=github_step_summary)
+    append_to_file(content=content, filepath=github_step_summary)
--- a/coverage_comment/github.py
+++ b/coverage_comment/github.py
@@ -195,3 +195,12 @@ def create_missing_coverage_annotation(annotation_type: str, file: str, line: in
         file=file,
         line=str(line),
     )
+
+
+def append_to_environment_file(content: str, filepath: pathlib.Path):
+    with filepath.open(mode="a") as file:
+        file.write(content)
+
+
+def add_job_summary(content: str, github_step_summary: pathlib.Path):
+    append_to_environment_file(content=content, filepath=github_step_summary)
--- a/coverage_comment/main.py
+++ b/coverage_comment/main.py
@@ -279,7 +279,7 @@ def save_coverage_data_files(
     markdown_report = coverage_module.generate_coverage_markdown()
 
     github.add_job_summary(
-        content=f"## Coverage report\n{markdown_report}",
+        content=f"## Coverage report\n\n{markdown_report}",
         github_step_summary=config.GITHUB_STEP_SUMMARY,
     )
 
--- a/coverage_comment/main.py
+++ b/coverage_comment/main.py
@@ -149,6 +149,11 @@ def generate_comment(
         return 1
 
     assert config.GITHUB_PR_NUMBER
+
+    github.add_job_summary(
+        content=comment, github_step_summary=config.GITHUB_STEP_SUMMARY
+    )
+
     try:
         if config.FORCE_WORKFLOW_RUN:
             raise github.CannotPostComment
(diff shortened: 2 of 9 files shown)


coverage_comment/github.py
```diff
@@ -195,3 +195,11 @@ def create_missing_coverage_annotation(annotation_type: str, file: str, line: in
         file=file,
         line=str(line),
     )
+
+
+def append_to_environment_file(content: str, filename: str):
+    print(f'"{content}" >> ${filename}')
```
ewjoachim, line 201: I bugged reading this line :D Indeed, WIP.

kieferro (commented):
I did a lot of trial and error to automatically verify that the summaries work for the e2e test as well. But I have not found a useful way. [There seems to be no API endpoint.](https://github.com/orgs/community/discussions/27649)

It is possible to see the raw summary text with an URL like https://github.com/org/repo/actions/runs/runid/jobs/jobid/summary_raw. However, you can only do that if you are logged in and have the appropriate permissions. So I tried to get the content with a token like [the method to get raw files from private repos](https://stackoverflow.com/questions/18126559/how-can-i-download-a-single-raw-file-from-a-private-github-repo-using-the-comman). But that doesn't seem to work either.

@ewjoachim If you have another idea, I'll be happy to incorporate it. But if not, I guess we'll have to do without.

coverage_comment/github.py
```diff
@@ -195,3 +195,12 @@ def create_missing_coverage_annotation(annotation_type: str, file: str, line: in
         file=file,
         line=str(line),
     )
+
+
+def append_to_environment_file(content: str, filepath: pathlib.Path):
+    with filepath.open(mode="a") as file:
+        file.write(content)
```
ewjoachim, line 202: Technically, this function doesn't care that this is an environment file, it only appends to a file.

coverage_comment/main.py
```diff
@@ -273,6 +278,11 @@ def save_coverage_data_files(
 
     markdown_report = coverage_module.generate_coverage_markdown()
 
+    github.add_job_summary(
+        content=f"## Coverage report\n{markdown_report}",
```
ewjoachim, line 282:
```suggestion
        content=f"## Coverage report\n\n{markdown_report}",
```
(I think I'm used to adding an empty line after titles in markdown)

tests/integration/test_github.py
```diff
@@ -332,3 +332,11 @@ def test_create_missing_coverage_annotation__annotation_type(capsys):
     assert (
         output.out.strip() == "::error file=test.py,line=42::This line has no coverage"
     )
+
+
+def test_add_job_summary(summary_file):
+    github.add_job_summary(
+        content="[job summary content]", github_step_summary=summary_file
+    )
```
ewjoachim, line 340: Can we call it twice to check that content was appended and not truncated ?

tests/integration/test_main.py
```diff
@@ -360,6 +366,10 @@ def test_action__push__default_branch(
 See more details and ready-to-copy-paste-markdown at:
     https://github.com/py-cov-action/foobar/tree/python-coverage-comment-action-data"""
     assert log == expected
+    assert "## Coverage report" in summary_file.read_text()
+    assert "Name" in summary_file.read_text()
+    assert "Stmts" in summary_file.read_text()
+    assert "Missing" in summary_file.read_text()
```
ewjoachim, line 372: (we could open and read the file once)

ewjoachim (approved):
Seems awesome ! 
Just a few comments.

tests/integration/test_github.py
kieferro: Yes, good idea

kieferro:
Thank you for the review, I've added the suggestions.
Do you need more time to test possibilities for the e2e tests or should we just not implement an automatic check there because there aren't any viable options?

ewjoachim:
> should we just not implement an automatic check there because there aren't any viable options

That.

At some point, we can't do GitHub's job for them.

ewjoachim (approved):
Let's merge !

kieferro:
> At some point, we can't do GitHub's job for them.

Yes, I completely agree

PR #207, opened 2023-07-12
Change order of steps to avoid wrong workflow command groups

kieferro:
I noticed that it can happen that the annotations for missing coverage are [mixed with other outputs](https://github.com/Aarhus-Psychiatry-Research/psycop-common/actions/runs/5241351155/jobs/9463329902#step:10:988). This is especially awkward because the annotations are grouped together in a collapsed view and therefore the other outputs can easily be overlooked.

I guess this is because the annotations use print and the other output uses the logging system. So I would suggest that we change the order to make sure that everything important is logged before the annotations are created.

I also tried this in a small test case and with the changes it seems to work better:
[Run with previous order](https://github.com/kieferro/test/actions/runs/5525347447/jobs/10078855464#step:7:987)
[Run with new order](https://github.com/kieferro/test/actions/runs/5525417755/jobs/10079008649#step:7:1096)


ewjoachim (approved):
Seems to make sense :) Not sure I understand 100% why the problem occured in the first place but as you said, probably a race condition in GitHub. And print vs log means stdout vs stderr.

ewjoachim:
(I believe changing pour print to `print([...], file=sys.stderr)` should be another way to fix this, but I'm ok with that fix, if you say it works as intended)

kieferro:
> (I believe changing pour print to `print([...], file=sys.stderr)` should be another way to fix this, but I'm ok with your fix, if you say it works as intended)

Good idea! I would prefer to use your suggestion as it is much cleaner because it means we don't have to worry about the order of output.
I committed the code for it. If it looks good to you, feel free to merge.

ewjoachim (approved):
I was wondering whether we were actually sure that workflow commands in stderr were understood by GitHub, but then I realized we wouldn't be having this issue if it was not.

PR #212, opened 2023-07-23
Create annotations based on diff-coverage, not full coverage

ewjoachim:
This is a revamp of https://github.com/py-cov-action/python-coverage-comment-action/pull/188

I believe that the part that failed was:
https://github.com/py-cov-action/python-coverage-comment-action/pull/188/files#diff-5327f4a45499e33cab8758f61e6b738fc200a07b00c58fc231f37a3142461ac7R194-R207

which I removed completely from this PR, as it wasn't needed.

1/ This PR being created by me, we'll get e2e tests
2/ I've remade the commits and remove the things are are much unrelated. We still have a 1st commit that could be somewhat elsewhere, but we were being really inconsistent in types used, and it was visible when switching between cover and diff-cover, so hopefully it's better now

### DIFF
--- a/coverage_comment/annotations.py
+++ b/coverage_comment/annotations.py
@@ -4,15 +4,19 @@ from coverage_comment import github
 MISSING_LINES_GROUP_TITLE = "Annotations of lines with missing coverage"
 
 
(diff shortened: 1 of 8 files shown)


kieferro (approved):
Thanks, that looks great

PR #213, opened 2023-07-23
Add SUBPROJECT_ID setting

### LINKED ISSUE #16: Multi-module monorepo support?
Soo, I spent quite some time trying to make the action work with our private monorepo setup and looks like I failed :-(

The problem is that when in monorepo, you most probably are running the pytest/coverage inside some sub-module. And thus even with relative files, the paths are relative to that submodule.

So, when being merged, you kinda lost the "submodule" info and now when running the action in the repo root, it rightfully complains about a missing .py file. The file is present, but just inside some of the sub-modules.

If there are any ideas/examples of monorepo setups, I'd be glad to see any suggestions.

Also, while the whole sub-module thing (at least in our case) has a root-level Poetry project, when running the same `coverage combine` or `coverage json` locally under a particular poetry shell, it works fine! I believe the reason for that is cause all the sub-modules are listed in the root project as dependencies and poetry resolves these quite well.

ewjoachim:
Fixes #16 
Needs #184 :heavy_check_mark: 
Renders #180 obsolete (sorry)

I've finally taken the time to redo my PR for allowing monorepo settings.

I'm introducing a single setting (`SUBPROJECT_ID`) that does multiple things:
- changes the marker in the comment to allow multiple comment not to overwrite one another
- changes the name of the file we write as output to avoid clashes
- changes the name of the data branch.

Of course, it will make a lot more sense after we merge #184 and add `COVERAGE_PATH` (though I believe there may be cases where we may want the 2 settings to be independant: it's not unimaginable to have the coverage run on the root for both projects. It's also reasonable to use COVERAGE_PATH outside of a monorepo setting (e.g. if using `src/`) so while the doc for `SUBPROJECT_ID` mentions `COVERAGE_PATH`, I don't believe the 2 settings will be linked in the code.

Full backwards compatibility to be expected.


ewjoachim:
Shall we ?

kieferro:
> Shall we ?

Oh, I'm sorry. I didn't realise that this was already review/merge ready. I thought you were still working on it because you never requested a review and because in some places the necessary changes have not yet been made. For example, [here](https://github.com/py-cov-action/python-coverage-comment-action/blob/d53fffc10f2bd40af385f492f2c400288c233d9e/coverage_comment/main.py#L172) and also at another place the used marker is not the formated, but still the old one (Or am I completely off track?). And I also noticed something else that I thought is still missing. So I'm sorry if you've been waiting for my review all this time.

ewjoachim:
Oh, don't worry, no it's probably me :) I'll self-review and ping you when it's ready !

ewjoachim:
I've re-read all and added tests.

kieferro (approved):
That looks great (:clap:). Very nice that we will soon be able to support monorepos with it. 
I've tested it fairly extensively in a test repo with multiple IDs in PR and push mode. And the code seems to run smoothly as far as I've seen.
I still had a few minor comments and I took the liberty of fixing the e2e tests that were failing.

I think it might be a good idea if we could somehow integrate the ID into `readme.md.j2` as well. Maybe even in the heading, so that the projects are easier to distinguish when looking at the README on the coverage data branch.

Otherwise, everything looks good :+1:

tests/end_to_end/conftest.py
```diff
@@ -243,7 +243,7 @@ def code_path(request):
 @pytest.fixture
 def subproject_id(request):
     mark = request.node.get_closest_marker("subproject_id")
-    return mark.args[0] if mark else None
```
ewjoachim: maybe it's the other way around: this fixture should return None, but where we use it, we should do `json.dumps(subproject_id)` (either are fine)

ewjoachim:
> I think it might be a good idea if we could somehow integrate the ID into readme.md.j2 as well. Maybe even in the heading, so that the projects are easier to distinguish when looking at the README on the coverage data branch.

Good idea !

tests/end_to_end/repo/.github/workflows/ci.yml
```diff
@@ -48,4 +48,4 @@ jobs:
         if: steps.coverage_comment.outputs.COMMENT_FILE_WRITTEN == 'true'
         with:
           name: python-coverage-comment-action
-          path: python-coverage-comment-action.txt
+          path: python-coverage-comment-action*.txt
```
ewjoachim, line 51: Note for self: check if we should do the same change in the README examples, if not already done.

tests/end_to_end/conftest.py
kieferro: Sure, I was just trying to get the tests working

tests/end_to_end/repo/.github/workflows/ci.yml
ewjoachim: No we're good, it was already part of the documentation.

ewjoachim:
I think we're good 🎉

kieferro:
Yes, looks great. Since we're both happy with it, I'll merge

PR #216, opened 2023-07-25
e2e tests on external PRs after owner approval

ewjoachim:
This PR adds a new mechanism to our "not-even-oh-so-slightly-overengineered" CI pipeline:
- When a PR is reviewed
- And it's an approval
- And the PR author is not a repo ~owner~ member
- And the approval author is a repo ~owner~ member
- EDIT: and the review body contains `/e2e`

Then we launch the end to end tests, on the commit that was approved

### DIFF
--- a/.github/workflows/e2e-external.yml
+++ b/.github/workflows/e2e-external.yml
@@ -15,6 +15,7 @@ jobs:
       github.event.pull_request.author_association != 'MEMBER'
       && github.event.review.author_association == 'MEMBER'
       && github.event.review.state == 'approved'
+      && contains(github.event.review.body, '/e2e')
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/.github/workflows/e2e-external.yml
+++ b/.github/workflows/e2e-external.yml
@@ -1,0 +1,56 @@
+name: End-to-end tests for external PRs
+
+on:
+  pull_request_review:
+    types: [submitted]
+
+concurrency:
+  group: ${{ github.event_name }}-${{ github.ref }}
+
+jobs:
+  test:
+    name: Run end to end tests
+    # If reviewed by a repo(/org) owner
+    if: |
+      github.event.pull_request.author_association != 'OWNER'
+      && github.event.review.author_association == 'OWNER
+      && github.event.review.state == 'approved'
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          # Important: use the commit that was reviewed. GitHub is making sure
+          # that this is race-condition-proof
+          ref: ${{ github.event.review.commit_id }}
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+
+      - name: Install Poetry
+        run: |
+          pipx install poetry --python=python3.10
+
+      - name: Poetry caches
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/
+          key: ${{ hashFiles('poetry.lock') }}
+
+      - name: Install deps
+        run: poetry install
+
+      - name: Run end-to-end tests
+        run: poetry run pytest tests/end_to_end
+        env:
+          COVERAGE_COMMENT_E2E_GITHUB_TOKEN_USER_1: ${{ secrets.COVERAGE_COMMENT_E2E_GITHUB_TOKEN_USER_1 }}
+          COVERAGE_COMMENT_E2E_GITHUB_TOKEN_USER_2: ${{ secrets.COVERAGE_COMMENT_E2E_GITHUB_TOKEN_USER_2 }}
+          COVERAGE_COMMENT_E2E_ACTION_REF: ${{ github.event.review.commit_id }}
+          COVERAGE_COMMENT_E2E_REPO_SUFFIX: ${{ github.event.pull_request.number }}
+          COVERAGE_COMMENT_E2E_PR_AUTHOR: ${{ github.event.pull_request.user.login }}
--- a/.github/workflows/e2e-external.yml
+++ b/.github/workflows/e2e-external.yml
@@ -12,8 +12,8 @@ jobs:
     name: Run end to end tests
     # If reviewed by a repo(/org) owner
     if: |
-      github.event.pull_request.author_association != 'OWNER'
-      && github.event.review.author_association == 'OWNER
+      github.event.pull_request.author_association != 'MEMBER'
+      && github.event.review.author_association == 'MEMBER'
       && github.event.review.state == 'approved'
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-external.yml
```diff
@@ -0,0 +1,55 @@
+name: End-to-end tests for external PRs
+
+on:
+  pull_request_review:
+    types: [submitted]
+
+concurrency:
+  group: ${{ github.event_name }}-${{ github.ref }}
```
ewjoachim, line 8: Here, `github.ref` [means](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_review):
> PR merge branch `refs/pull/:prNumber/merge`

ewjoachim:
Ok, we're actually both "MEMBER", not owners

.github/workflows/e2e-external.yml
```diff
@@ -0,0 +1,56 @@
...
+    name: Run end to end tests
+    # If reviewed by a repo(/org) owner
+    if: |
+      github.event.pull_request.author_association != 'MEMBER'
+      && github.event.review.author_association == 'MEMBER'
+      && github.event.review.state == 'approved'
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      contents: write
```
ewjoachim, line 21:
```suggestion
```

ewjoachim:
I see one remaining problem that I don't know how to solve and if it's worth solving: if we both approve a PR, e2e tests will run twice.

.github/workflows/e2e-external.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: poetry install
+
+      - name: Run end-to-end tests
+        run: poetry run pytest tests/end_to_end
+        env:
+          COVERAGE_COMMENT_E2E_GITHUB_TOKEN_USER_1: ${{ secrets.COVERAGE_COMMENT_E2E_GITHUB_TOKEN_USER_1 }}
+          COVERAGE_COMMENT_E2E_GITHUB_TOKEN_USER_2: ${{ secrets.COVERAGE_COMMENT_E2E_GITHUB_TOKEN_USER_2 }}
+          COVERAGE_COMMENT_E2E_ACTION_REF: ${{ github.event.review.commit_id }}
+          COVERAGE_COMMENT_E2E_REPO_SUFFIX: ${{ github.event.pull_request.number }}
+          COVERAGE_COMMENT_E2E_PR_AUTHOR: ${{ github.event.pull_request.user.login }}
```
kieferro, line 53:
```suggestion
      - name: Check cache if the tests have already run
        id: cache
        uses: actions/cache/restore@v3
        with:
          key: ${{ github.ref }}-${{ github.event.review.commit_id }}
          path: dummy.txt

      - name: Checkout
        if: steps.cache.outputs.cache-hit != 'true'
        uses: actions/checkout@v3
        with:
          # Important: use the commit that was reviewed. GitHub is making sure
          # that this is race-condition-proof
          ref: ${{ github.event.review.commit_id }}

      - name: Set up Python
        if: steps.cache.outputs.cache-hit != 'true'
        uses: actions/setup-python@v4
        with:
          python-version: "3.10"

      - name: Install Poetry
        if: steps.cache.outputs.cache-hit != 'true'
        run: |
          pipx install poetry --python=python3.10

      - name: Poetry caches
        if: steps.cache.outputs.cache-hit != 'true'
        uses: actions/cache@v3
        with:
          path: |
            ~/.cache/
          key: ${{ hashFiles('poetry.lock') }}

      - name: Install deps
        if: steps.cache.outputs.cache-hit != 'true'
        run: poetry install

      - name: Run end-to-end tests
        if: steps.cache.outputs.cache-hit != 'true'
        run: poetry run pytest tests/end_to_end
        env:
          COVERAGE_COMMENT_E2E_GITHUB_TOKEN_USER_1: ${{ secrets.COVERAGE_COMMENT_E2E_GITHUB_TOKEN_USER_1 }}
          COVERAGE_COMMENT_E2E_GITHUB_TOKEN_USER_2: ${{ secrets.COVERAGE_COMMENT_E2E_GITHUB_TOKEN_USER_2 }}
          COVERAGE_COMMENT_E2E_ACTION_REF: ${{ github.event.review.commit_id }}
          COVERAGE_COMMENT_E2E_REPO_SUFFIX: ${{ github.event.pull_request.number }}
          COVERAGE_COMMENT_E2E_PR_AUTHOR: ${{ github.event.pull_request.user.login }}

      - name: Create dummy file
        if: steps.cache.outputs.cache-hit != 'true'
        run: touch dummy.txt
        shell: bash

      - name: Create cache to indicate that the tests have run
        if: steps.cache.outputs.cache-hit != 'true'
        uses: actions/cache/save@v3
        with:
          key: ${{ github.ref }}-${{ github.event.review.commit_id }}
          path: dummy.txt
```

> I see one remaining problem that I don't know how to solve and if it's worth solving: if we both approve a PR, e2e tests will run twice.

I think the easiest way to make this work is to use the cache system and to create a cache for the commit hash when the tests run. Whether it's worth it I don't know either, but at least it wouldn't add so much complexity.

.github/workflows/e2e-external.yml
ewjoachim: If you approve, and then the test fail, and then I approve, I believe the code you suggested will change the result to "Ok", because the tests won't run, but the steps will be succesful.

.github/workflows/e2e-external.yml
ewjoachim: I could be wrong, but I think if:
- you approve a PR
- tests are red
- then I approve the PR

Then the step is going to run with everything skipped, so it will succeed, and turn the check green (honnestly, I'm not even entirely sure that there will be a check displayed on the UI. I'd like to say there will but we'll need to test this)

kieferro: > Then the step is going to run with everything skipped

If one step of the workflow fails, won't all subsequent steps be canceled? That was actually my reasoning behind this. Since the cache is added in the very last step, this should only happen if all steps before (i.e. also the tests) were successful.

ewjoachim: Ok, I've tested and:
- The result of the workflow _do_ appear in the PR checks
- But the last run overrides the state, whether this be a success, a failure of even the whole step skipped with an if

I'm going to add something easier: e2e tests will only launch if the review body contains /e2e

ewjoachim:
Okay: [Only launch tests if review body contains /e2e](https://github.com/py-cov-action/python-coverage-comment-action/pull/216/commits/82243efad7e080470248495708c464e6ed9d814c)

This also gives us an easy way to relaunch them if need be.

.github/workflows/e2e-external.yml
kieferro: > I'm going to add something easier: e2e tests will only launch if the review body contains /e2e

Sure, that works too

ewjoachim:
Ok, this will only start working once it's merged to the default branch. So if you're ok, let's merge !

kieferro (approved):
Thanks, looks good. Ready to merge when you are

PR #217, opened 2023-07-28
Upgrade to Python 3.11

ewjoachim:
There are 2 commits. I think I'll need to isolate the 1st commit into its own branch, then merge it to let the CI create the v2 image, and then launch the CI for the rest of the PR and test everything. I wanted your opinion before doing anything too harsh @kieferro . Of course, this can wait until after the other ongoing PRs.


kieferro (approved):
Just by looking at it, that seems reasonable. And if we merge the first commit (which shouldn't really breaking anything). Then we can also see for the second one if all tests are passing.

ewjoachim:
I've cheated and manually pushed the v2 to docker. So if the CI is green, I guess we should be all good.

ewjoachim:
Tests are failing because we're matching on an exception and py3.11 exceptions are improved :)

Dockerfile.build
```diff
@@ -15,7 +15,6 @@ RUN git config --system --add safe.directory '*'
 WORKDIR /workdir
 
 COPY pyproject.toml ./
-COPY poetry.lock ./
```
kieferro, line 18: Just for my understanding, why don't we use the lockfile here anymore?

ewjoachim: Ah yes, I forgot to add a line about that. I believe that it's not used when we do a `pip install .`. If my understanding of the specifics of poetry is correct, it's used when we do `poetry install` which installs the dev env, but when we do `pip install`, then the only thing that matters is the dependency specifications from pyproject.toml (https://github.com/py-cov-action/python-coverage-comment-action/blob/v3/pyproject.toml#L18-L23) which only specify wildcards. I thought it was a bit confusing to include an unused file.

But I can be wrong !

ewjoachim:
(small test, ignore me)

Dockerfile.build
kieferro: Thanks for the explanation, that makes sense of course.

ewjoachim:
All good then ?

kieferro:
Yes, looks good @ewjoachim

PR #219, opened 2023-07-30
Instead of deleting e2e repo when tests succeed, delete them when PR closes

ewjoachim:
I believe this should make debugging easier. Also, I'm working on a revamp of the dev env, trying to replicate locally the e2e tests, because when they fail it's often tedious...


tests/end_to_end/conftest.py
```diff
@@ -276,19 +276,15 @@ def f(*args):
 
     yield f
 
-    if not is_failed():
-        gh_delete_repo(gh_me)
```
kieferro, line 280: Using a deletion workflow when the PR is closed seems like a good idea to me. But I actually thought it was quite nice to be invited only if the tests fail. What's the idea behind changing that?

ewjoachim: It's open to discussions. If you do a run and you want to eyeball the changes in the public repo, just to see your changes, you can. For private repo, keeping the previous behaviour would mean you can't check the changes unless they failed the test.

Maybe we should add (yet) another workflow command for inviting everyone. We have the pattern now, it would be easy. Like: you'd send `/invite` and it would invite people (you, me and the PR author).

kieferro: Yeah, thanks, I think that's the best way

kieferro (approved):

PR #231, opened 2023-08-05
Switch to pyproject.toml where applicable

ewjoachim:
(I'm also wondering about switching to ruff, and I'm getting the feeling that mypy isn't very helpful the way it's set up, but that's a smaller-scoped start)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,38 +1,5 @@
-[isort]
-profile = black
-skip = .venv,.tox
-
 [flake8]
 # E203: whitespace before colon on list slice: mylist[1 : 2]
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -1,38 +1,5 @@
-[isort]
-profile = black
-skip = .venv,.tox
-
 [flake8]
 # E203: whitespace before colon on list slice: mylist[1 : 2]
 # E501: line too long (black knows better)
 extend-ignore = E203,E501
 extend-exclude = .venv
```
ewjoachim, line 5: https://github.com/PyCQA/flake8/issues/234

PR #237, opened 2023-08-12
Redo versionning

### LINKED ISSUE #233: Version 3 tag does not point to the correct location
I noticed discrepancies with the version tag and the branch `v3`. I assume you where working in the branch `v3` to prep for that version, but the tag `v3.0.0` points to [this Dockerfile @v3.0.0.](https://github.com/py-cov-action/python-coverage-comment-action/blob/v3.0.0/Dockerfile) that has a link to the image `@v1`, which was not the intent I assume?

What happens now if a user uses this action is confusing:
uses: py-cov-action/python-coverage-comment-action@v3 the **branch** will be checked out, and Docker image `:v3` will be used.
uses: py-cov-action/python-coverage-comment-action@v3.0.0 the **tag** will be checked out, and Docker image `:v1` will be used.

I'd recommend cleaning up the branches after merging them to main, and correct the `v3` tag to point to the right commit. It's best to keep things straightforward and use tags everywhere, and not branches.

ewjoachim:
Fixes #233 

- Switch to `main` branch
- On each commit to `main`, after tests pass, (re-)tag the action as `vx`, `vx.y` and `vx.y.z`

When this is merged, it's probably worth doing a `v3.1.0`. Also, we need to change the default branch and update the existing PRs, and delete the v3 branch

At this point I have no idea how to handle minor or major versions (e.g. how to make it so when we merge a PR that warrants a version change, it's not immediately released as incremental)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ on:
   workflow_run:
     workflows: [CI]
     types: [completed]
-    branches: [v3]
+    branches: [main]
 
 concurrency:
   group: release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,3 +34,33 @@ jobs:
           tags: ewjoachim/python-coverage-comment-action-base:v3
           push: true
           file: Dockerfile.build
+
+  compute-tags:
+    name: Re-tag action with new version
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'push' }}
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Apply new tags
+        run: |
+          git fetch --tags origin
+          current="$(git describe --tags --abbrev=0 --match 'v*.*.*')"
+
+          tag_major="$(echo $current | cut -d. -f1)"
+          git tag -f ${tag_major}
+          git push -f origin ${tag_major}
+
+          tag_minor="$(echo $current | cut -d. -f1,2)"
+          git tag -f ${tag_minor}
+          git push -f origin ${tag_minor}
+
+          increment=$(echo $current | cut -d. -f3)
+          increment=$((increment + 1))
+          tag_incremental="${tag_minor}.${increment}"
+          git tag ${tag_incremental}
+          git push origin ${tag_incremental}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,8 +11,11 @@ on:
     types: [completed]
     branches: [v3]
 
+concurrency:
+  group: release
+
 jobs:
-  push_to_registry:
+  push-to-registry:
     name: Push Docker image to Docker Hub
     runs-on: ubuntu-latest
     # Check if the tests were successful and were launched by a push event
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,23 +49,16 @@ jobs:
       - name: Apply new tags
         run: |
           git fetch --tags origin
-          current="$(git describe --tags --abbrev=0 --match 'v*.*.*')"
-          should_bump_minor=$(pipx run conventional list-commits --from-last-tag --parse | jq --slurp 'any(.[]; .data.subject.breaking == "!")')
+          current="$(git describe --tags --abbrev=0 --match 'v*.*')"
 
           major="$(echo $current | cut -d. -f1)"
           minor="$(echo $current | cut -d. -f2)"
-          patch="$(echo $current | cut -d. -f3)"
 
           git tag -f ${major}
           git push -f origin ${major}
 
           # Major releases will be released manually.
-          if [ "${should_bump_minor}" = "true" ]; then
-            minor=$((minor + 1))
-            new_tag="${major}.${minor}.0"
-          else
-            patch=$((patch + 1))
-            new_tag="${major}.${minor}.${patch}"
-          fi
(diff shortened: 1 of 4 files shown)


README.md
kieferro, line 1: We should also change [this](https://github.com/py-cov-action/python-coverage-comment-action/blob/versionning/README.md?plain=1#L348) link to point to `main` instead of `v3`

.github/workflows/release.yml
```diff
@@ -0,0 +1,66 @@
...
+  compute-tags:
+    name: Re-tag action with new version
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'push' }}
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
kieferro, line 47:
```suggestion
        uses: actions/checkout@v3
```

kieferro (commented):
> At this point I have no idea how to handle minor or major versions (e.g. how to make it so when we merge a PR that warrants a version change, it's not immediately released as incremental)

Yes, this is a point that I also noticed

We should also make sure that we set a very current tag before deleting the branch, so that the action doesn't fall back to the v3.0.0 tag for a short time because no new tag has been created yet, but also no branch exists anymore.

kieferro:
I thought about it a bit more and somehow this versioning system doesn't feel convincing to me. We create a new tag for each commit (for which the CI is running) and I don't quite see how our workflow should work when we want to increase the major or minor version. And I'm not so convinced about pushing tags around that may be linked to official releases either.

What if we do it a bit differently and use the following system:
- After the CI runs, we delete all tags ending with `-development`.
- Then we take the last tag with schema `v*.*.*` and add a `-development` after it and use that as the tag for the latest commit.
- If we want to increase the major or minor version, we just tag the latest commit with the new version number

So the basic idea would be that we have a development tag that we push around so that everyone who just uses `@v3` is always up to date and the "real" version tags always stay in the same place.

That might also have disadvantages that I'm not thinking of right now, but that would somehow appeal to me as a system better than moving the "real" release tags around or creating a whole lot of tags.

ewjoachim:
Well, what I suggested is what GitHub says we should do, unless I'm mistaken:

> We recommend creating releases using semantically versioned tags – for example, v1.1.3 – and keeping major (v1) and minor (v1.1) tags current to the latest appropriate commit. For more information, see "About custom actions" and "About semantic versioning.
[source](https://docs.github.com/en/actions/creating-actions/releasing-and-maintaining-actions)

The only change I did was adding minor tags too but I can remove that. The part where we force-push tags is their idea.

That said, we can discuss alternatives for sure!

Can you expand on the benefit from your solution compared to having people use `@main`? Using `@vx.y.z-development` in your workflow will only work until we release the next incremental which feels a bit like a pain, no ? Plus dependabot is likely to be confused and suggest to people to switch from dev to non-dev tag, though I can't say for sure.

ewjoachim:
Another alternative could be to just say we release manually once in a while (which we haven't been doing a lot) and have our workflow repush the `vX` tag when we push a release. This means people on `v3` will get the latest release instead of the bleeding edge, so it may take a while to figure out if there's a problem.

.github/workflows/release.yml
```diff
@@ -0,0 +1,71 @@
...
+          tags: ewjoachim/python-coverage-comment-action-base:v3
+          push: true
+          file: Dockerfile.build
+
+  compute-tags:
+    name: Re-tag action with new version
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'push' }}
+    permissions:
+      contents: write
```
ewjoachim, line 43: we may miss some permissions here.

ewjoachim: No we're good. `contents: write` includes releases: https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents

ewjoachim:
I just thought of 2 other possibilities:

- we keep a 2-number version instead of a 3 number. So there's just minor and major. We do a minor tag every commit on main, major tags are handled by hand. Or:
- [calver](https://calver.org). But given we don't plan releases in advance, that might get weird.

kieferro:
> Can you expand on the benefit from your solution compared to having people use `@main`?

Compared to that, there is really no advantage. If we were to come up with a concept for an action that no one uses yet, I would absolutely be in favour of people who want bleeding-edge functions simply using `@main`, and we would keep manually publishing a release every once in a while without force pushing the tags, so that everyone else gradually gets the new functions as well.
But the problem is that pretty much all the people who use the action are already using `@v3`. I also thought that was the problem we were trying to solve. Namely, that everyone who uses `@v3` should continue to get bleeding-edge. That was the intention behind my suggestion, that we find a way between force pushing tags and having everyone switch to `@main`, by effectively making `@v3` an `@main` by always adding a tag at the most recent commit that redirects the `@v3` to an `@main`.

ewjoachim:
But then I'm not sure what usecase the "development" tags would help ?

ewjoachim:
(Switched to `vx.y` without incremental. This eliminates the need for distinguishing incremental releases from minor ones. If we're releasing with every PR and always keeping backwards compat anyway, then there's no real difference)

kieferro:
> But then I'm not sure what usecase the "development" tags would help ?

Ah no, my mistake, I'm sorry. I made some wrong assumptions and thought that Github would have more logic for finding a specific version. Just ignore what I suggested.

kieferro (approved):

PR #248, opened 2023-09-03
Run PR activities on push on non-default branch

### LINKED ISSUE #234: Check for open PRs on `push` GHA trigger
Hello 👋 

I've been successfully using this action for a couple weeks now (it is great 👏 ) - I am opening this issue to suggest an improvement:

When the action is run as part of a `push` it would be great if the action could lookup the open PR that the commit belongs to and post the PR comment there.

I think this could simplify the usage of this action by obviating the need for:
- `pull_request` trigger
- `actions/upload-artifact@v3`
- `workflow_run` trigger

(I am not suggesting that the above functionality is removed, only suggesting that we improve the `push` event handling for now, which could make this action easier to use imo)

What do you think? If this sounds like an acceptable improvement I would also be willing to work on it 🙂

### LINKED ISSUE #246: Switch to new note syntax
Syntax for admonitions, including `Note` has changed: https://github.com/orgs/community/discussions/16925#discussioncomment-6506860

> A new syntax, `[!NOTE]`, has been added, which will gradually replace the old one. However, the old syntax will continue to work for some time.

We're using the old syntax here:

https://github.com/py-cov-action/python-coverage-comment-action/blob/16ccd40b7d40599a9b69ebe1d8558c37a0f740ea/coverage_comment/template_files/comment.md.j2#L18

Let's replace it.

(This may be made redundant by #211 but I'm just opening the issue to make sur we don't forget it if need be)

ewjoachim:
Closes #234, Closes #246

Still needs:
- Manual tests
- A decision regarding whether we need e2e tests for that or not
- Documentation (especially on how to write the part that posts the comment from the artifact. Though it will be very much like CommentPR.yml from [here](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/))

But the code is written, and we already have 100% coverage so that's a good start.

### DIFF
--- a/coverage_comment/main.py
+++ b/coverage_comment/main.py
@@ -64,7 +64,7 @@ def action(
     event_name = config.GITHUB_EVENT_NAME
     if event_name not in {"pull_request", "push", "workflow_run"}:
         log.error(
-            'This action has only been designed to work for "pull_request", "branch" '
+            'This action has only been designed to work for "pull_request", "push" '
             f'or "workflow_run" actions, not "{event_name}". Because there are security '
             "implications. If you have a different usecase, please open an issue, "
             "we'll be glad to add compatibility."
--- a/coverage_comment/main.py
+++ b/coverage_comment/main.py
@@ -119,11 +119,21 @@ def process_pr(
         base_ref=base_ref, coverage_path=config.COVERAGE_PATH
     )
 
-    previous_coverage_data_file = storage.get_datafile_contents(
-        github=gh,
-        repository=config.GITHUB_REPOSITORY,
-        branch=config.COVERAGE_DATA_BRANCH,
-    )
+    # It only really makes sense to display a comparison with the previous
+    # coverage if the PR target is the branch in which the coverage data is
+    # stored, e.g. the default branch.
+    # In the case we're running on a branch without a PR yet, we can't know
+    # if it's going to target the default branch, so we display it.
+    previous_coverage_data_file = None
+    pr_targets_default_branch = base_ref == repo_info.default_branch
+
+    if pr_targets_default_branch:
+        previous_coverage_data_file = storage.get_datafile_contents(
+            github=gh,
+            repository=config.GITHUB_REPOSITORY,
+            branch=config.COVERAGE_DATA_BRANCH,
+        )
+
     previous_coverage = None
     if previous_coverage_data_file:
         previous_coverage = files.parse_datafile(contents=previous_coverage_data_file)
@@ -135,6 +145,7 @@ def process_pr(
             previous_coverage_rate=previous_coverage,
             base_template=template.read_template_file("comment.md.j2"),
             custom_template=config.COMMENT_TEMPLATE,
+            pr_targets_default_branch=pr_targets_default_branch,
         )
     except template.MissingMarker:
         log.error(
--- a/coverage_comment/main.py
+++ b/coverage_comment/main.py
@@ -119,11 +119,6 @@ def process_pr(
         base_ref=base_ref, coverage_path=config.COVERAGE_PATH
     )
 
-    if config.ANNOTATE_MISSING_LINES:
-        annotations.create_pr_annotations(
-            annotation_type=config.ANNOTATION_TYPE, diff_coverage=diff_coverage
-        )
-
     previous_coverage_data_file = storage.get_datafile_contents(
         github=gh,
         repository=config.GITHUB_REPOSITORY,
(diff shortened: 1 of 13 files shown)


coverage_comment/main.py
```diff
@@ -60,62 +61,68 @@ def action(
...
+    )
+    base_ref = config.GITHUB_BASE_REF or repo_info.default_branch
+    diff_coverage = coverage_module.get_diff_coverage_info(
+        base_ref=base_ref, coverage_path=config.COVERAGE_PATH
+    )
+
+    if config.ANNOTATE_MISSING_LINES:
+        annotations.create_pr_annotations(
+            annotation_type=config.ANNOTATION_TYPE, diff_coverage=diff_coverage
+        )
```
kieferro, line 125:
```suggestion
```
I think it would make sense to move this step a bit so that it is not executed when we have a push event for which there is no PR yet and it is therefore just unnecessary output. (See comment below for where we could move it)

coverage_comment/main.py
```diff
@@ -151,21 +158,35 @@ def generate_comment(
...
+        try:
+            pr_number = github.find_pr_for_branch(
+                github=gh,
+                # A push event cannot be initiated from a forked repository
+                repository=config.GITHUB_REPOSITORY,
+                owner=config.GITHUB_REPOSITORY.split("/")[0],
+                branch=config.GITHUB_BRANCH_NAME,
+            )
+        except github.CannotDeterminePR:
+            pr_number = None
```
kieferro, line 179:
```suggestion
        except github.CannotDeterminePR:
            pr_number = None
        else:
            if config.ANNOTATE_MISSING_LINES:
                annotations.create_pr_annotations(
                    annotation_type=config.ANNOTATION_TYPE, diff_coverage=diff_coverage
                )
```
(See above)

coverage_comment/main.py
kieferro, line 1: A problem that we had before, but which I noticed during the review and which we can perhaps change relatively easily with this PR: We don't check whether the target branch of the PR is the default branch. This means that, theoretically, a comment is also generated for PRs that are opened between two (non-default) branches, but this comment is then based on the coverage rate of the default branch and therefore makes no sense.
We could also change the documentation in the README and change the triggers depending on the target branch. But I think it would make sense to also include it in the implementation to also cover cases in which the action has been set up in a wrong way or before those changes of the documentation

kieferro (commented):
Great, looks good :+1:. I still have a few small comments, but I tested with a copy of the template repo and everything seems to work fine.

coverage_comment/main.py
ewjoachim: > this comment is then based on the coverage rate of the default branch and therefore makes no sense

Hm, there are 2 distinct things:

Coverage comparison with default branch. I'd tend to say this makes sense even for PR not targeting the default branch, but you're right that in some cases it may not. It may make sense in the context of an epic where we'd make a branch for the epic, merge multiple PRs and then merge the branch. It may not make sense when making a hotfix PR on an old release.

Diff-coverage: restricting coverage calculations to lines added by this PR. This supposedly uses the PR head and not the default branch as comparison point (except when not available) especially to avoid the case of a PR not targeting the default branch.

With this in mind, can you expand a bit in terms of what would make no sense, so I'm certain we're talking about the same thing? And just to understand, you're suggesting we  would not issue the comment at all if we're not targeting the default branch, right?

Also, I'm not sure we can ask the users to change their workflow triggers just for us: that will impact their whole CI. I guess we could ask them to add an `if` clause to the action, but it makes more sense to deal with that solely from within our code. After all, it's a problem affecting a very small percentage of PRs, and the only issue it brings is comments that may not be 100% relevant but that still contain some accurate info.

kieferro: > With this in mind, can you expand a bit in terms of what would make no sense, so I'm certain we're talking about the same thing

I was referring to the `previous_coverage_rate`, which we obtain from the coverage data branch and which therefore always refers to the default branch, and from my point of view this does not make much sense in cases where the PR doesn't target the default branch. Diff coverage, for example, is something different because it actually compares the two relevant branches. But if the comment says "Coverage went from X to Y" and X does not refer to the target, but to the default branch, I think that only creates confusion.



> And just to understand, you're suggesting we would not issue the comment at all if we're not targeting the default branch, right?

Not necessarily, but at least until we have sorted out how to handle this case.

ewjoachim:
Comments addressed. For the question of the default branch, I've:
- removed previous coverage data when running on a PR not targeting the default branch
- but kept it in case we're running on a branch with no PR yet: we can't know.

ewjoachim:
I've also changed the `Note` that displays where there's no reference coverage info to use a specific message in the case it's because the PR doesn't target the default branch

kieferro (approved):
Looks great :+1:
> A decision regarding whether we need e2e tests for that or not

I'm not quite sure if this is so easy to test if we also want to test the normal use cases and therefore want to keep `pull_request` as a trigger. We could build some system that this trigger is ignored for example if the PR title contains a certain word and we create a PR which contains that word and tests if it also works only through the push trigger, but I don't know if that's worth the effort. What's your assessment?

ewjoachim:
> What's your assessment?

I agree with you :)

PR #251, opened 2023-09-05
Split release into "automated" (@v3) and "manual" (@v3.x)

ewjoachim:
As discussed with @kieferro : there will be a manual button for releasing when we want to release. The @v3 tag will be automatically re-pushed after every commit.

### DIFF
--- a/.github/workflows/manual-release.yml
+++ b/.github/workflows/manual-release.yml
@@ -1,0 +1,35 @@
+name: Release Action (manual)
+
+on: workflow_dispatch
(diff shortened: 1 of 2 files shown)


kieferro (approved):
Thanks, looks good

ewjoachim:
https://github.com/py-cov-action/python-coverage-comment-action/actions/workflows/manual-release.yml (Getting there from the repo and clicking is 4 clicks, but it's your choice :D )

PR #255, opened 2023-09-09
Also store raw coverage information on data branch

kieferro:
Prepares #211

With the current redesign for the new coverage comment, we need more information about the coverage on the default branch.
The easiest way to make sure that we have all the data we need, is to simply store the raw JSON coverage data on the coverage data branch.
This is not used yet, but the sooner we ship this change, the more repos will have a push event between this change and the merge of #211, which leads to more cases where all data is available.

### DIFF
--- a/coverage_comment/coverage.py
+++ b/coverage_comment/coverage.py
@@ -67,13 +67,15 @@ def compute_coverage(num_covered: int, num_total: int) -> decimal.Decimal:
     return decimal.Decimal(num_covered) / decimal.Decimal(num_total)
 
 
-def get_coverage_info(merge: bool, coverage_path: pathlib.Path) -> Coverage:
+def get_coverage_info(
+    merge: bool, coverage_path: pathlib.Path
+) -> tuple[dict, Coverage]:
     try:
         if merge:
             subprocess.run("coverage", "combine", path=coverage_path)
 
-        json_coverage = subprocess.run(
-            "coverage", "json", "-o", "-", path=coverage_path
+        json_coverage = json.loads(
+            subprocess.run("coverage", "json", "-o", "-", path=coverage_path)
         )
     except subprocess.SubProcessError as exc:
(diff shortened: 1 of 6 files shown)


coverage_comment/coverage.py
```diff
@@ -89,7 +91,7 @@ def get_coverage_info(merge: bool, coverage_path: pathlib.Path) -> Coverage:
             )
         raise
 
-    return extract_info(data=json.loads(json_coverage), coverage_path=coverage_path)
+    return json_coverage, extract_info(data=json_coverage, coverage_path=coverage_path)
```
ewjoachim, line 94: I'm always hesitant to returning tuples. Sure, destructuring is great, but I'm always afraid of mixing up the order of elements. That said, risks are low here, so probably not worth the hassle of changing it.

coverage_comment/coverage.py
```diff
@@ -67,13 +67,15 @@ def compute_coverage(num_covered: int, num_total: int) -> decimal.Decimal:
...
+def get_coverage_info(
+    merge: bool, coverage_path: pathlib.Path
+) -> tuple[dict, Coverage]:
     try:
         if merge:
             subprocess.run("coverage", "combine", path=coverage_path)
 
-        json_coverage = subprocess.run(
-            "coverage", "json", "-o", "-", path=coverage_path
+        json_coverage = json.loads(
```
ewjoachim, line 77: (I tend to try and limit things that happen in a try statement to the statements that may raise. But it's not very important, as in this case, it's clear the json.loads can't interact with the subprocess error.)

ewjoachim (approved):
Nice ! Let's merge this if it's all good with you !

coverage_comment/coverage.py
kieferro: Yes, I wasn't entirely happy with that either, but with type annotations and CI we should also notice very quickly if we swap the two elements.

PR #256, opened 2023-09-10
Small release system improvements

kieferro:
I noticed two things regarding the releases:
* We always use the action version from one push before for our own CI because the `v3` force-push is done after our CI runs. I would suggest that we always use the very latest version instead, which generally makes sense as we know what we change, but also results in the v3 tag not being pushed if our own usage of the action fails in the CI.
* Release notes are created relative to the latest tag, which in most cases is v3, but we want them relative to the latest minor version. (I fixed this for [v3.9](https://github.com/py-cov-action/python-coverage-comment-action/releases/tag/v3.9) by hand)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
 
       - name: Coverage comment
         id: coverage_comment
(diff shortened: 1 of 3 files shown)


ewjoachim (approved):
Great idea !

PR #257, opened 2023-09-10
Remove diff cover

ewjoachim:
This PR removes diff-cover from the codebase:
- Computes the list of added lines from `git diff`
- Creates a DiffCoverage object from a Coverage object, restricting lines to the ones added
- Recomputes all derived metrics from the base ones

The nice thing is that the tests barely need to change.

We bump the docker image version number in the process, but I believe the old action will run on the new image, and vice versa (thanks to the `pip install .` that we left in `Dockerfile`) so I don't think we need to do anything specific for this to work.

### DIFF
--- a/coverage_comment/coverage.py
+++ b/coverage_comment/coverage.py
@@ -3,6 +3,7 @@ import datetime
 import decimal
 import json
 import pathlib
+from collections.abc import Iterable
 
 from coverage_comment import log, subprocess
 
@@ -42,6 +43,12 @@ class Coverage:
     meta: CoverageMetadata
     info: CoverageInfo
     files: dict[pathlib.Path, FileCoverage]
+
+
+# The format for Diff Coverage objects may seem a little weird, because it
+# was originally copied from diff-cover schema. In order to keep the
+# compatibility for existing custom template, we kept the same format.
+# Maybe in v4, we can change it to a simpler format.
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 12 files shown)


ewjoachim:
Ha of course.
The CI of e2e fail because:
> ERROR: failed to solve: ewjoachim/python-coverage-comment-action-base:v4: docker.io/ewjoachim/python-coverage-comment-action-base:v4: not found

ewjoachim:
```
subprocess.CalledProcessError: Command '('git', 'diff', '--unified=0', 'origin/main', '--', '.')' returned non-zero exit status 128.
fatal: bad revision 'origin/main'
```
Hmm... Is it a problem of a missing fetch ? But why didn't we have that with diff cover ? Time to explore...

ewjoachim:
https://github.com/py-cov-action/python-coverage-comment-action/pull/257/files#diff-5327f4a45499e33cab8758f61e6b738fc200a07b00c58fc231f37a3142461ac7L215 That's the reason.

ewjoachim:
Ow damn you already have a branch for doing that :( https://github.com/py-cov-action/python-coverage-comment-action/tree/replace-diff-cover

Sorry about the duplicated work !

kieferro:
> Sorry about the duplicated work !

It's of course not ideal, but no big problem, I had neglected the branch a bit anyway, because we had more important things at the time and and then I didn't deal with it. But the fact that I had already thought about it makes the review process easier for me.

How do we want to proceed, do you want to get the CI passing first or should I review it now?

ewjoachim:
I think the only thing needed is for me to push the missing git fetch. Let's say you're welcome to review once it's green :)

ewjoachim:
All green, rebased conflicts resolved, all yours :)

coverage_comment/coverage.py
```diff
@@ -211,57 +217,94 @@ def extract_info(data: dict, coverage_path: pathlib.Path) -> Coverage:
...
+            continue
+
+        executed = set(file.executed_lines) & set(added_lines_for_file)
+        count_executed = len(executed)
+
+        missing = set(file.missing_lines) & set(added_lines_for_file)
+        count_missing = len(missing)
+        # We don't want to count twice the lines that are both executed and
+        # missing (e.g. the branch lines partially executed)
+        count_total = len(executed | missing)
```
kieferro, line 243: I may well be wrong about that, but I don't think this is necessary. To my knowledge lines are never in both `executed_lines` and `missing_lines` and they indicate the branches [via `executed_branches` and `missing_branches`](https://github.com/nedbat/coveragepy/blob/afa75596342d2463a5676486885d1b0b489e20c0/coverage/jsonreport.py#L114-L119).

coverage_comment/coverage.py
```diff
@@ -211,57 +217,94 @@ def extract_info(data: dict, coverage_path: pathlib.Path) -> Coverage:
...
+
+        total_num_lines += count_total
+        total_num_violations += count_missing
+
+        percent_covered = decimal.Decimal("1")
+        if count_total != 0:
+            # You can't multiply a decimal by a float but you can divide it
+            # by an int, that's why we split the following into 2 lines
+            percent_covered *= count_executed
+            percent_covered /= count_total
```
kieferro, line 253: If it is indeed the case that there are lines that occur in both `executed` and `missing`, then we have to change the calculation system here. For example, if `executed` and `missing` contained exactly the same lines, because they are branches that are always only half executed, then something around 50% should come out. But since `count_total` only takes each line into account once, the result would be 100%.

tests/unit/test_coverage.py
```diff
@@ -114,3 +114,119 @@ def test_generate_coverage_markdown(mocker):
...
+-foo
+-bar
++foo1
++bar1
++foo2
++bar2
+--- a/foo.txt
++++ b/foo.txt
+@@ -0,0 +1 @@
++bar
```
kieferro, line 225:
```suggestion
+bar
@@ -8 +7,0 @@
-foo
```
I think it would also be useful to test for deletions without any additions

tests/unit/test_coverage.py
kieferro, line 1: Perhaps we could test the `get_diff_coverage_info` function in more detail, which we didn't have to do so far due to the external dependency. I would find a test with several files quite good, for example to check that the calculation of the total values also works with several files.
We could, for example, use `coverage_obj` and add one more file and otherwise design the test similar to the one that already exists.

coverage_comment/coverage.py
```diff
@@ -211,57 +217,94 @@ def extract_info(data: dict, coverage_path: pathlib.Path) -> Coverage:
...
+
+    if total_num_lines + total_num_violations != 0:
+        final_percentage *= total_num_lines - total_num_violations
+        final_percentage /= total_num_lines
+
+    return DiffCoverage(
+        total_num_lines=total_num_lines,
+        total_num_violations=total_num_violations,
+        total_percent_covered=final_percentage,
+        num_changed_lines=num_changed_lines,
```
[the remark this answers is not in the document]
ewjoachim: The problem is we don't know if people out there have been using custom templates that use it :(

coverage_comment/coverage.py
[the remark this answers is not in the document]
ewjoachim: Absolutely :o

tests/unit/test_coverage.py
ewjoachim: Good idea !

coverage_comment/coverage.py
ewjoachim: You're probably right, but I guess the easiest way to find out is to test :)

ewjoachim: You're right :)
```console
$ tree
.
├── coverage.json
├── my.py
└── test_x.py

1 directory, 3 files

$ cat my.py
import os


def x():
    if os.environ.get("A"):
        print("A")
    print("B")

$ cat test_x.py
from my import x


def test_x():
    x()

$ pytest --cov-report=json --cov my --cov-branch
====================================================== test session starts =======================================================
platform darwin -- Python 3.9.18, pytest-7.4.2, pluggy-1.3.0
rootdir: /private/tmp/plop
plugins: cov-4.1.0
collected 1 item

test_x.py .                                                                                                                [100%]

---------- coverage: platform darwin, python 3.9.18-final-0 ----------
Coverage JSON written to file coverage.json


======================================================= 1 passed in 0.12s ========================================================

$ cat coverage.json | jq '.files."my.py" | .executed_lines, .missing_lines'
[
  1,
  4,
  5,
  7
]
[
  6
]
```

coverage_comment/coverage.py
ewjoachim: Thankfully, you were right and lines only appear once.

coverage_comment/coverage.py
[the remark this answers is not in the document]
kieferro: Right, good point, I hadn't thought about that

ewjoachim:
Ok, I think all your comments have been addressed.

kieferro (approved):
Yes, looks good

PR #263, opened 2023-09-13
Restore missing commits

kieferro:
Some commits seem to have been lost due to the change of the default branch and this PR adds these commits back.
So far this is just a copy of #213 but there may be more commits we need to recover.
There were also some Renovate-PRs that went missing like #241 and #240 but Renovate already took care of that in #245

(@ewjoachim I've sent you a message with some more details)


ewjoachim:
I've checked that there doesn't seem to be other commits missing.

https://github.com/py-cov-action/python-coverage-comment-action/assets/1457576/fc0f58dd-126b-4575-80c9-dccc2dbc3431

ewjoachim (approved):

PR #271, opened 2023-09-21
Store `COVERAGE_PATH` on data branch

kieferro:
Addition to #255
Working on #211 I realized that we also need to store the coverage path to be able to compare the branches

### DIFF
--- a/coverage_comment/files.py
+++ b/coverage_comment/files.py
@@ -59,6 +59,7 @@ class ReplaceDir:
 def compute_files(
     line_rate: decimal.Decimal,
     raw_coverage_data: dict,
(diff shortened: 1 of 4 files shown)


ewjoachim:
Is it to support cases where the coverage path changed between the main branch and the PR ?

ewjoachim (approved):
not sure yet I understand why but the code does what it says it does.

kieferro:
> Is it to support cases where the coverage path changed between the main branch and the PR ?

Yes. This is of course a fairly rare case, but it may be that all the files remain in the same place, but the place from which they are tested changes.

PR #328, opened 2023-12-24
Speed-up builds

### LINKED ISSUE #268: Instructions for running built docker container.
Hey, thanks for the action!

I'm trying to figure out how to use the build docker container version to save build times, but it doesnt seem to be a drop in replacement:

```
- name: Coverage Report
      if: inputs.skip-coverage == 'false'
      id: coverage_comment
      uses: docker://docker.io/ewjoachim/python-coverage-comment-action:v2
      with:
          GITHUB_TOKEN: ${{ inputs.gh-token }}
          COVERAGE_PATH: ${{ inputs.project-root }}
 ```

Results in:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/coverage_comment/settings.py", line 167, in from_environ
    config_obj = cls(**config)
                 ^^^^^^^^^^^^^
TypeError: Config.__init__() missing 2 required keyword-only arguments: 'GITHUB_TOKEN' and 'GITHUB_PR_RUN_ID'
During handling of the above exception, another exception occurred:
```

So maybe I need to provide different config parameters for the built docker image version..

ewjoachim:
Closes #268 

This PR adds a test when building Dockerfile at the start of the action that only runs `pip install` if pyproject.toml has changed since the image was built.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,7 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           # See Dockerfile.build for instructions on bumping this.
-          tags: ewjoachim/python-coverage-comment-action-base:v4
+          tags: ewjoachim/python-coverage-comment-action-base:v5
           push: true
           file: Dockerfile.build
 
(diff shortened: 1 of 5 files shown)


ewjoachim:
To test this, I'll need to push v5 from my laptop. So I'll need to build v5 locally. Now that my laptop is an ARM, I'll need to learn how to cross-build docker images for x64. Seems feasible but... For another day.

Issue #375, opened 2024-02-29
Support branch coverage


ewjoachim:
See https://github.com/py-cov-action/python-coverage-comment-action/pull/361#issuecomment-1954220360

olivier-lacroix:
Supporting branch coverage would be great !

ewjoachim:
Why is this ticket still open ?

We had https://github.com/py-cov-action/python-coverage-comment-action/pull/466 in the meantime, which I believe solved the issue ? @olivier-lacroix are there still things missing ? (I'm closing the ticket, but if there are, I'll reopen)

olivier-lacroix:
Ah! Great to know! Having seen this open issue, I had not activated it. Pretty sure all will be well , thanks!

PR #413, opened 2024-05-26
Branch coverage

ewjoachim:
(just some WIP I had lying around, for supporting branch coverage, #375 )

### DIFF
--- a/coverage_comment/coverage.py
+++ b/coverage_comment/coverage.py
@@ -6,6 +6,7 @@ import decimal
 import json
 import pathlib
 from collections.abc import Sequence
+from typing import cast
 
 from coverage_comment import log, subprocess
 
@@ -40,6 +41,8 @@ class FileCoverage:
     executed_lines: list[int]
     missing_lines: list[int]
     excluded_lines: list[int]
+    executed_branches: list[list[int]] | None
+    missing_branches: list[list[int]] | None
     info: CoverageInfo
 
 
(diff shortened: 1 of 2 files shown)


coverage_comment/coverage.py
```diff
@@ -82,10 +85,26 @@ class DiffCoverage:
...
+    use_branch_coverage: bool,
+    num_branches_covered: int | None,
+    num_branches_total: int | None,
+) -> decimal.Decimal:
+    """Compute the coverage percentage, with or without branch coverage."""
+    if use_branch_coverage:
+        num_branches_covered = cast(int, num_branches_covered)
+        num_branches_total = cast(int, num_branches_total)
+        numerator = decimal.Decimal(num_covered + num_branches_covered)
+        denominator = decimal.Decimal(num_total + num_branches_total)
```
ferdnyc, line 100: This math exactly checks out, with the `raw_data.totals` numbers and `raw_data.totals.percent_covered` produced by pytest.

coverage_comment/coverage.py
```diff
@@ -82,10 +85,26 @@ class DiffCoverage:
...
 
 
-def compute_coverage(num_covered: int, num_total: int) -> decimal.Decimal:
-    if num_total == 0:
+def compute_coverage(
+    num_covered: int,
+    num_total: int,
+    use_branch_coverage: bool,
+    num_branches_covered: int | None,
+    num_branches_total: int | None,
```
ferdnyc, line 93: Rather than conditionally including the branch numbers, you could just arrange to have them default to `0` if there's no branch data. If they're both `0` they won't affect the math, so they can just be added in unconditionally.

ewjoachim:
Closing in favor of #466

PR #424, opened 2024-06-06
Improve checks on github.ref fields

ollaw:
This PR arises from the problem of having `"pull"` substring inside your branch name.

In particular, having a branch name like `ISSUE-XXX/task/enable-pull-lorem-ipsum`  causes a crash in the action since  the if inside `GITHUB_PR_NUMBER` property wrongly assumes that `github.ref` refers to a PR, and tries to cast to integer the PR number. 

Goal of this PR is to make the check a bit stronger, for both `GITHUB_PR_NUMBER` and `GITHUB_BRANCH_NAME` (since it's easier to incur in same problem with a tags that contains `"heads"` substring inside it)  checking `github.ref` head rather than randomly checking that a word is inside it.

### DIFF
--- a/coverage_comment/settings.py
+++ b/coverage_comment/settings.py
@@ -125,14 +125,14 @@ class Config:
     @property
     def GITHUB_PR_NUMBER(self) -> int | None:
         # "refs/pull/2/merge"
-        if "pull" in self.GITHUB_REF:
+        if self.GITHUB_REF.startswith("refs/pull"):
             return int(self.GITHUB_REF.split("/")[2])
         return None
 
     @property
     def GITHUB_BRANCH_NAME(self) -> str | None:
         # "refs/heads/my_branch_name"
-        if "heads" in self.GITHUB_REF:
+        if self.GITHUB_REF.startswith("refs/heads"):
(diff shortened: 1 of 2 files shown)


ewjoachim (commented):
Damn, of course :facepalm: 

Thanks for the contribution, would you mind adding a couple of tests in https://github.com/py-cov-action/python-coverage-comment-action/blob/main/tests/unit/test_settings.py#L126-L145 to ensure we don't get a regression ? Thanks !

ewjoachim:
I think I never thought it would happen because I've mainly tested with `on: pull_request` whereas I guess you're experiencing this with `on: push` ?

ollaw:
> whereas I guess you're experiencing this with on: push ?

Exactly, I also unfairly blamed GtiHub at first, I thought there was some problem on push / pull_request triggers 😄 

> would you mind adding a couple of tests [...] ?

Sure, take a look if that's okay.

ewjoachim (approved):
Perfect 🤩

ewjoachim:
https://github.com/py-cov-action/python-coverage-comment-action/releases/tag/v3.24

Issue #442, opened 2024-07-18
Job summary shows truncated message


tgehrs:
We have a pull request that exceeds the max_files number and the python-coverage-comment-action displays this message at the bottom of the coverage comment.

<img width="862" alt="Screenshot 2024-07-18 at 2 19 47 PM" src="https://github.com/user-attachments/assets/199ddb90-74fd-4407-94a2-beb656ce3776">

When I navigate to the workflow summary page, it shows a similar message about truncating:

`https://github.com/<org>/<repo>/actions/runs/<run_id>/attempts/<attempt_num>#summary-<summary_id>`

<img width="923" alt="Screenshot 2024-07-18 at 2 20 10 PM" src="https://github.com/user-attachments/assets/f1d84bca-26f3-4dc4-8dd4-19b2c8f56590">

From what I can tell, the add_job_summary function is passed the same comment as was generated for the pull request, which includes the default max_files:

https://github.com/py-cov-action/python-coverage-comment-action/blob/e1188d571bebf0a4db96a5fba886b537a0f90457/coverage_comment/main.py#L167-L184

https://github.com/py-cov-action/python-coverage-comment-action/blob/e1188d571bebf0a4db96a5fba886b537a0f90457/coverage_comment/main.py#L202-L204

I would be happy to create a pull request to resolve this if needed, I would propose to either re-generate the markdown without the max_files parameter or conditionally re-generate it if the number of files exceeds the max_files setting. Is there a preference from the authors/maintainers?

Also, it might be nifty to add a link to the workflow summary from the error message, I could try to take a crack at that too as part of this fix.

ewjoachim:
Ah oops, I believe I completely forgot to generate the full report, you're right.

> I would propose to either re-generate the markdown without the max_files parameter or conditionally re-generate it if the number of files exceeds the max_files setting. Is there a preference from the authors/maintainers?

I believe the generation is likely a few milliseconds, so either way, but likely generate it twice will lead to the clearest code.

> Also, it might be nifty to add a link to the workflow summary from the error message, I could try to take a crack at that too as part of this fix.

Good idea!

tgehrs:
Closed with #444

PR #444, opened 2024-07-19
add summary comment

tgehrs:
Fix for #442 to create a separate summary comment from the truncated PR comment

### DIFF
--- a/coverage_comment/main.py
+++ b/coverage_comment/main.py
@@ -189,6 +189,7 @@ def process_pr(
             previous_coverage_rate=previous_coverage_rate,
             files=files_info,
             count_files=count_files,
+            max_files=None,
             minimum_green=config.MINIMUM_GREEN,
             minimum_orange=config.MINIMUM_ORANGE,
             repo_name=config.GITHUB_REPOSITORY,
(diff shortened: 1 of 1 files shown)


coverage_comment/main.py
```diff
@@ -182,6 +182,24 @@ def process_pr(
             marker=marker,
             subproject_id=config.SUBPROJECT_ID,
         )
+        summary_comment = template.get_comment_markdown(
```
ewjoachim, line 185:
```suggestion
        # Same as above except `max_files` is None
        summary_comment = template.get_comment_markdown(
```

ewjoachim:
Can you look at existing tests and tell me if it's easy to add a new one for this case ?

tgehrs:
From my reading of the tests, this one seems to confirm that this functionality should work. It seems that mocks are rarely used, so I'm not sure how else to verify that it is called with None for max_files. If you have a place in mind you would like me to add a test please let me know and I will gladly do my best.

https://github.com/py-cov-action/python-coverage-comment-action/blob/fafc0aee90b594ca6eb2c4689a137b9ea19b6731/tests/unit/test_template.py#L327-L357

ewjoachim:
(Ah sorry, missed you message).

I agree with you, adding a test there will probably mean more mocks, and limited value. Let's go !

ewjoachim (approved):

ewjoachim:
Released in https://github.com/py-cov-action/python-coverage-comment-action/releases/tag/v3.26 :)

PR #453, opened 2024-08-04
add pagination for artifacts retrieval

fabiogallotti:
Hello @ewjoachim,

following the discussion we had some weeks ago with @AlessandroMiola [here](https://github.com/py-cov-action/python-coverage-comment-action/discussions/438), I'm proposing you a small fix in order to handle the pagination for artifacts.

I'm looping page by page and, if the result of page n+1 is empty, I'm breaking out of the loop.
The rationale behind is that If we ask a page without artifact we get an empty response (see for example https://api.github.com/repos/pydantic/pydantic/actions/runs/9815232181/artifacts?page=3)

To be honest, I'm not very happy with this solution, probably the best one would be to migrate from the https://github.com/michaelliao/githubpy client (that I noticed it's not maintained anymore) to something like https://github.com/PyGithub/PyGithub, but the changes would be substantial I guess.

I also took the occasion to introduce a Makefile to ease local development, let me know what you think about it ;) 

Closes #438

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@
 .coverage
 dev-env-vars
 .coverage*
+.python-version
+__pycache__
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,4 @@
 .coverage
 dev-env-vars
 .coverage*
-.python-version
 __pycache__
-.vscode
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,21 @@
+.DEFAULT_GOAL := help
+SHELL = bash
+
+
+.PHONY: help
+help: ## show this help
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+.PHONY: install
+install: ## install dependencies
+	poetry install --with dev
+	poetry run pre-commit install
+
+.PHONY: lint
+lint: ## lint code
+	poetry run ruff check --fix coverage_comment tests
+	poetry run ruff format coverage_comment tests
+
+.PHONY: test
+test: ## run all tests
+	poetry run pytest tests -vv --cov-report term-missing --cov=coverage_comment
--- a/Makefile
+++ b/Makefile
@@ -1,21 +1,12 @@
-.DEFAULT_GOAL := help
-SHELL = bash
-
-
-.PHONY: help
-help: ## show this help
-	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
-
 .PHONY: install
 install: ## install dependencies
 	poetry install --with dev
-	poetry run pre-commit install
+	pre-commit install
 
 .PHONY: lint
 lint: ## lint code
-	poetry run ruff check --fix coverage_comment tests
-	poetry run ruff format coverage_comment tests
+	pre-commit
 
 .PHONY: test
 test: ## run all tests
-	poetry run pytest tests -vv --cov-report term-missing --cov=coverage_comment
+	poetry run pytest tests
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -0,0 +1,21 @@
...
+	poetry run pre-commit install
+
+.PHONY: lint
+lint: ## lint code
+	poetry run ruff check --fix coverage_comment tests
+	poetry run ruff format coverage_comment tests
+
+.PHONY: test
+test: ## run all tests
+	poetry run pytest tests -vv --cov-report term-missing --cov=coverage_comment
```
ewjoachim, line 21: I believe all those arguments are already in pyproject.toml, no need to repeat them here.

Makefile
```diff
@@ -0,0 +1,21 @@
...
+
+.PHONY: install
+install: ## install dependencies
+	poetry install --with dev
+	poetry run pre-commit install
+
+.PHONY: lint
+lint: ## lint code
+	poetry run ruff check --fix coverage_comment tests
+	poetry run ruff format coverage_comment tests
```
ewjoachim, line 17: Pre-commit takes care of linting. A more appropriate (and more complete) lint command would be
```console
pre-commit run --all-files
```
Or even just
```console
pre-commit
```

Makefile
```diff
@@ -0,0 +1,21 @@
...
+
+
+.PHONY: help
+help: ## show this help
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+.PHONY: install
+install: ## install dependencies
+	poetry install --with dev
+	poetry run pre-commit install
```
ewjoachim, line 12: pre-commit shouldn't be installed inside poetry, so that's rather `pre-commit install` without `poetry run`

Makefile
ewjoachim, line 1: I'm not a huge fan of Makefiles, mainly because they tend to look like a mess (not commenting your Makefile in particular, but it's an inherent property of the format). Also, they lost me many years ago when I discovered you have to mix tabs and space on some lines.

My philosophy is more to try and use standard tools and configure them so they're ready to use.

In the case of your Makefile: if you think the project is missing some directions on how to launch some actions, you're right of course, but I'd like to direct you to changing the CONTRIBUTING.md doc first. That said, I'm totally ok to keep the Makefile as a documentation TL;DR, but let's make it so (remove the help target, fix the few comments I left which should show that targets are synonym to a 1 (or few) words command.

If you're curious, my go-to approach these days when I need to make complex commands accessible in a project is a `scripts` folder, which offers the advantage of letting you write some scripts with Python when they become too complex to be readable in bash. But I don't think we need this here. A Makefile will be very OK.

coverage_comment/github.py
```diff
@@ -54,7 +54,14 @@ def download_artifact(
...
+    page = 1
+    artifacts = []
+    while True:
+        result = repo_path.actions.runs(run_id).artifacts.get(page=str(page))
+        if not result:
+            break
+        artifacts.extend(result.artifacts)
+        page += 1
     try:
         artifact = next(
```
ewjoachim, line 66: What about incorporating reading the pages and paginating into one generator (`yield`), so that if you find the artifact on the first page, you don't need to fetch the second page ? Also, this will limit the number of tests you have to change while still letting you do dedicated tests for cancellation.

Also, if the number of elements github returns on a page is fewer than the max, you know in advance there won't be a next page,

ewjoachim (commented):
Thank you very much for tackling this ! We're off to a good start, though I believe we could improve some parts.

I'm also discussing the Makefile decision in the comments below.

Thank you again. If at any point, you'd rather I take it from here, let me know :)

.gitignore
```diff
@@ -3,3 +3,6 @@
 .coverage
 dev-env-vars
 .coverage*
+.python-version
+__pycache__
+.vscode
```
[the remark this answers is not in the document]
fabiogallotti: Oh, thanks for letting me know this

coverage_comment/github.py
fabiogallotti: you're right, I'll work on improving it

fabiogallotti: To avoid unnecessarily fetching the next page, I leveraged on the `total_count` attribute of the response, instead of using the github default max page size of 30, it seems safer to me

ewjoachim (approved):
Awesome ! Thanks for your first contribution to this project, and for your patience !

PR #466, opened 2024-08-31
Branch coverage

ferdnyc:
This PR builds on @ewjoachim 's WIP #413, completing the changes and updating the relevant tests so that all tests pass successfully when run locally with `poetry run pytest`. The pre-commit hooks were also run on all commits, and any ruff changes incorporated. I have not, however, run the E2E tests; I figured that could wait until  after the PR was opened.

The PR branch contains four commits, the first being @ewjoachim's WIP changes to `coverage_comment.coverage`. The second commit is my own changes to complete the branch-coverage support.
 
A single new unit test, `test_compute_coverage_with_branches`, is added to `tests/unit/test_coverage.py` in the third commit, to exercise the updated `coverage_comment.coverage.compute_coverage` functionality in cases where (non-zero) branch coverage values are included.

The fourth commit adjusts the `conftest.py` fixtures and the various tests in `test_template.py` to both supply and expect coverage data computed accounting for branch coverage, raising the actual coverage for the fixture(s) data from 60% to <s>53.84%</s> 62.5%.

I'll add a bit of commentary on specific changes as review comments, so they're easier to discuss in the proper context.


coverage_comment/coverage.py
```diff
@@ -82,10 +85,20 @@ class DiffCoverage:
...
+    num_covered: int,
+    num_total: int,
+    num_branches_covered: int = 0,
+    num_branches_total: int = 0,
+) -> decimal.Decimal:
+    """Compute the coverage percentage, with or without branch coverage."""
+    num_branches_covered = cast(int, num_branches_covered)
+    num_branches_total = cast(int, num_branches_total)
+    numerator = decimal.Decimal(num_covered + num_branches_covered)
+    denominator = decimal.Decimal(num_total + num_branches_total)
```
ferdnyc, line 98: I decided to modify `compute_coverage` so that branch data is always included in the computations. The numbers now default to `0` (instead of `None`) if branch data isn't being tracked, so the math comes out right no matter what. If branch data is collected, it's factored in to the computations. If not, then `0` gets added to various values, and the numbers come out exactly the same either way.

coverage_comment/coverage.py
```diff
@@ -138,6 +151,26 @@ def generate_coverage_markdown(coverage_path: pathlib.Path) -> str:
...
+            num_branches_covered=data.get("covered_branches", 0),
+            num_branches_total=data.get("num_branches", 0),
+        ),
+        missing_lines=data["missing_lines"],
+        excluded_lines=data["excluded_lines"],
+        num_branches=data.get("num_branches"),
+        num_partial_branches=data.get("num_partial_branches"),
+        covered_branches=data.get("covered_branches", 0),
+        missing_branches=data.get("missing_branches", 0),
+    )
```
ferdnyc, line 171: I broke this out from `extract_info` (and probably should've done so in a separate commit, sorry) to avoid repeating basically the same construction twice with different arguments, and also to cut down on the indentation which was getting a bit out of hand.

coverage_comment/coverage.py
```diff
@@ -41,6 +42,8 @@ class FileCoverage:
     missing_lines: list[int]
     excluded_lines: list[int]
     info: CoverageInfo
+    executed_branches: list[list[int]] | None = None
+    missing_branches: list[list[int]] | None = None
```
ferdnyc, line 46: Giving these default values meant moving them _after_ the `info` element, in the `DataClass`. Hopefully that won't cause any problems? The code seems to always use keyword arguments when constructing them.

coverage_comment/coverage.py
```diff
@@ -284,18 +262,9 @@ def get_diff_coverage_info(
...
-        if coverage.meta.branch_coverage:
-            num_branches_covered = 12
-            num_branches_total = 12
-
         percent_covered = compute_coverage(
             num_covered=count_executed,
             num_total=count_total,
-            use_branch_coverage=coverage.meta.branch_coverage,
-            num_branches_covered=num_branches_covered,
-            num_branches_total=num_branches_total,
```
ferdnyc: I chose _not_ to use the branch data when computing _diff_ coverage even if it's available, which seems like it might be correct. To be honest, I'm not 100% on the theory there. But I noticed that even with branch coverage, each line in the _diff_ will either appear in the `covered_lines` set or the `missing_lines` set. I suppose if a diff line appears in `covered_lines`, but only has partial coverage because it appears in both `executed_branches` and `missing_branches`, then that might adjust the diff numbers down, but I figure a followup PR can add branch-coverage support to file diffs if necessary.

ewjoachim:
> I have not, however, run the E2E tests; I figured that could wait until after the PR was opened.

You're right. It's extremely annoying to run e2e tests for external contributors, and it's perfectly ok to wait for maintainers to launch them. Sadly, we need to use 2 github accounts to run those, but, at least, they provide a real-world test that's much better to trust the changes.

coverage_comment/coverage.py
```diff
@@ -255,8 +284,18 @@ def get_diff_coverage_info(
         total_num_lines += count_total
         total_num_violations += count_missing
 
+        num_branches_covered = None
+        num_branches_total = None
+        if coverage.meta.branch_coverage:
+            num_branches_covered = 12
+            num_branches_total = 12
```
ewjoachim: lol 12?

coverage_comment/coverage.py
ewjoachim: I hate positional arguments when there's more than 1 argument.

Also: the rule that arguments with default must follow positional arguments only holds when arguments may be positional. By indicating to Python that arguments must be named (with `*`), this rule disappears:

```python
def a(b=1, c):  # Syntax error
    ...
def a(*, b=1, c):  # Valid
    ...
a(2, 3)  # TypeError because arguments following `*` MUST be named
```
In this case, I suggest you add a `*` before the first argument, and we'll get a proper exception in case I used positional arguments :)

EDIT: oops, I just noticed it's a dataclass, not a function signature. In that case:
```python
@dataclass(kw_only=True)
```

coverage_comment/coverage.py
```diff
@@ -88,20 +88,14 @@ class DiffCoverage:
...
-        num_branches_covered = cast(int, num_branches_covered)
-        num_branches_total = cast(int, num_branches_total)
-        numerator = decimal.Decimal(num_covered + num_branches_covered)
-        denominator = decimal.Decimal(num_total + num_branches_total)
-    else:
-        numerator = decimal.Decimal(num_covered)
-        denominator = decimal.Decimal(num_total)
-
+    num_branches_covered = cast(int, num_branches_covered)
+    num_branches_total = cast(int, num_branches_total)
```
ewjoachim, line 96: Cast is unneeded now :) (and pre-commit will tell you but the import of cast at the top of the file is likely unneeded too, then)

Honnestly, having to use `cast` is a code smell. The simple fact that you're able to remove it is proof enough that it's the right decision :)

But I'll even go further: if we want to make sure that branch coverage is supported everywhere, maybe `num_branches_covered` and `num_branches_total` should not be optional.

coverage_comment/coverage.py
ewjoachim: (following my previous comment: let's actually make those mandatory so the problem disappears)
(but you can keep the `kw_only`, it doesn't hurt)

coverage_comment/coverage.py
ewjoachim: definitely a good choice.

coverage_comment/coverage.py
ewjoachim: branch coverage + diff coverage = 🤯 

At some point, understanding exactly what the number you're looking at means becomes so complex that's it likely not useful nor usable. I think it's the right call.

(it's already hard enough to not confuse the Diff coverage as the coverage rate of the added lines, vs the diff of the coverage, as in how the coverage rate evolved between before and now.)

tests/conftest.py
```diff
@@ -282,10 +282,10 @@ def _(code: str, has_branches: bool = True) -> coverage_module.Coverage:
...
                 missing_lines=0,
                 excluded_lines=0,
-                num_branches=0 if has_branches else None,
-                num_partial_branches=0 if has_branches else None,
-                covered_branches=0 if has_branches else None,
-                missing_branches=0 if has_branches else None,
+                num_branches=0,
+                num_partial_branches=0,
+                covered_branches=0,
+                missing_branches=0,
```
ewjoachim, line 288: (this screams that it was the right call)

coverage_comment/coverage.py
```diff
@@ -138,6 +151,26 @@ def generate_coverage_markdown(coverage_path: pathlib.Path) -> str:
...
+        percent_covered=compute_coverage(
+            num_covered=data["covered_lines"],
+            num_total=data["num_statements"],
+            num_branches_covered=data.get("covered_branches", 0),
+            num_branches_total=data.get("num_branches", 0),
+        ),
+        missing_lines=data["missing_lines"],
+        excluded_lines=data["excluded_lines"],
+        num_branches=data.get("num_branches"),
+        num_partial_branches=data.get("num_partial_branches"),
```
ferdnyc, line 168: And I'm now realizing I should've defaulted these to `0` as well.

```suggestion
        num_branches=data.get("num_branches", 0),
        num_partial_branches=data.get("num_partial_branches", 0),
```

ewjoachim (commented):
I'll dig a bit but I'm surprised we don't need to update the template.

For now, very nice start. Feel free to say if you have some feedback on the codebase & contribution process, especially if you saw things that were improvable, confusing, concerning etc.

Of course, this applies to the next steps of your contribution too.

ewjoachim (approved):
/e2e

(approving just to run the e2e tests for now)

coverage_comment/coverage.py
ewjoachim: (It's my code, so just to make it clear, I'm lmao to my own shortcomings. Your contribution is very good in comparison :D )

coverage_comment/coverage.py
ewjoachim: If you've realized this now, then I'd say this deserves a unit test :)

ferdnyc:
> branch coverage + diff coverage = 🤯

That was precisely my reaction as well, at which point I went: "Maybe I'm overthinking it. Even with branches, lines in the _diff_ are either covered or not, right?" I don't actually know if that IS right, but I figured I'd run with it and see how it goes. #YOLO

tests/conftest.py
```diff
@@ -461,7 +461,7 @@ def coverage_json():
         "totals": {
             "covered_lines": 6,
             "num_statements": 10,
-            "percent_covered": 60.0,
+            "percent_covered": 53.84615384615384615384615385,
```
ferdnyc, line 464: I'm more than a little concerned that _this_ number may shift a bit based on OS, CPU type, Python version, etc... we may want to finesse the fixture data so the figures come out slightly rounder.

ewjoachim:
I think the reason the e2e test don't seem to appear in the checks is because the branch is named `branch_coverage` and GitHub is confused with my older branch named the same. That said, they ran [here](https://github.com/py-cov-action/python-coverage-comment-action/actions/runs/10646711986/job/29513860832) producing:
- a public [repo](https://github.com/mihcaojwe/python-coverage-comment-action-end-to-end-466-public)
- and a private repo (link in a sec)


/invite

tests/conftest.py
ferdnyc: (Then again, since it's just `0.538461` repeating, maybe it won't be a problem.)

ewjoachim: You're probably right, IEEE 754 is a rabbit whole I'd rather not dig.

coverage_comment/coverage.py
ferdnyc: > But I'll even go further: if we want to make sure that branch coverage is supported everywhere, maybe `num_branches_covered` and `num_branches_total` should not be optional.

Well, we _could_, but since I'm not using it in the diff computations, defaulting it here means I can call `compute_coverage()` **there** without having to supply explicit `0`s, which I kind of liked. So it goes both ways.

I also liked that I didn't have to change the `test_compute_coverage()` code at all, just add a separate `test_compute_coverage_with_branches()` that supplies branch values.

But I'm open to either, it's not a hill I'm looking to die on.

ewjoachim: Nah, I'm ok with an opinion sufficiently backed by good arguments. It's not like it's a decision we can't revisit later either so let's go your way.

ewjoachim:
Do we need to update some of the numbers & wording in the template ?

ewjoachim:
So just to follow:
- [ ] we may want to adjust the exact value to get a coverage rate that doesn't have infinite decimals (https://github.com/py-cov-action/python-coverage-comment-action/pull/466#pullrequestreview-2273987292)
- [ ] Add a unit test [there](https://github.com/py-cov-action/python-coverage-comment-action/pull/466#pullrequestreview-2273964572)
- [ ] Regarding the comment template: the wording may be very slightly off because it says "statement" where it would actually now be "statement or branch". I'm not sure we want to overcomplicate things, so it's ok if we leave it this way, but if you want to update it, it's welcome :)

ferdnyc:
@ewjoachim 

> Do we need to update some of the numbers & wording in the template ?

That's a good question — there was one spot where, in the `template_test`, I ended up changing a statement like "the coverage is 60% (6/10)" to "the coverage is 53.84% (6/10)" which struck me as more than a little suspicious. So it's probably worth figuring out a better way to express the values there.

Regarding other changes, obviously things like the unit test addition and any template changes would be new commits, but for adjustments to existing changes (like the `@dataclass(kw_only=True)` tweak), do you prefer separate additional commits for reviewability, or rewritten and force-pushed PR commits for a clean history?

ferdnyc:
@ewjoachim Oh, also, can I get an invite to the private e2e repo?

ewjoachim:
> @ewjoachim Oh, also, can I get an invite to the private e2e repo?

Oh, interesting ! It should have been done automatically and it didn't
https://github.com/py-cov-action/python-coverage-comment-action/actions/runs/10646793964/job/29514024275?pr=466

Looks like it's because `true == 'true'` evaluates to `false` in GHA lingo

Let's fix it (in another PR)

ewjoachim:
> do you prefer separate additional commits for reviewability, or rewritten and force-pushed PR commits for a clean history?

_TL;DR: as you with I don't have strong opinions. But I appreciate discussing it and get to know what you prefer._

Glad you asked !
It's perfectly ok to rebase your commits for the sake of commit cleanliness, but it's true that it makes reviewing slightly harder. Nowadays, though, GitHub tries to help you a bit but it's not always ideal.
It's also perfectly ok if the commit history resembles what really happened in the contribution. I haven't spent a lot of time reading the commit history on this project, so I'm not 100% sure this is the best way to spend precious contributing time, but you seem interested :)

It sounds like you might be introduced to the wonderful of `git commit --fixup` (if you don't happen to know already).

When you wish to modify a commit but don't want to hinder reviewability, make the changes that would supposedly go in a commit and commit them with `git commit --fixup <hash of that commit>`. This will create a commit named `!fixup <name of the commit you targeted`. Push this for review.

When everything has been reviewed and you're ready to merge, use `git rebase --autosquash` (compatible with `--interactive`) and those commits will be automatically moved in `fixup` mode, right after the commit they target (in other words, they will be merged into this commit) and voilà :) (of course, you'll need to push force for this.

Of course, there are also cons to this approach (some of which are shared with other solutions):
- it requires more manual operations
- when I'm ready to merge, either I'll need to rebase your commits, or wait for you to do so
- The risk exists that I'll forget and commit the fixup commits which is a dent in the history
- Also, there's a theoretical risk that you'll have a conflict when rebasing, and that the conflict is not handled properly leading to issues, especially though I'll only do a light review on the final commits
- And of course, there's a risk that you're a malicious actor and you'll embed malicious content when you rebase your fixups, hoping that I won't see it as I don't have a reason to suspect the post-rebase review will be relevant.

> [!NOTE]
> I mentioned force pushing, and I'm _contractually required_ to mention using `--force-with-lease` instead or `--force`. If you already know about it, feel free to skip this note. Otherwise, here you are. `--force-with-lease` works the same as `--force` as long as no one else pushed to your branch. In the event that someone pushed to your branch: either you've already fetched their commits (they're already in your `origin/<branch>` ref) and git considers you know about them and knowingly wishes to overwrite them, or there are commits on the branch that you don't know about (not yet in your `origin/<branch>`) and git won't let you push force (you need to fetch first, and possibly deal with those commits if you want). In other words, when you push force after a rebase, `--force-with-lease` is always the same or better than `--force`

ewjoachim:
Let's retry after the fix.

/invite

ewjoachim:
It worked https://github.com/py-cov-action/python-coverage-comment-action/actions/runs/10657091714/job/29536514522?pr=466

You should have received the invitation :)

ferdnyc:
Finally had a chance to return to this. Not complete yet, but changes so far:

* I rewrote the commit updating the test data, so that the _new_ test data works out to a cleaner (and more rational) 62.5% coverage.
* Added new commits to:
  - Make the complex dataclasses `kw_only`
  - Dump the casting in `compute_coverage`
  - Fix the defaulting of branch stats in `_make_coverage_info()`

The last one still needs that unit test, and I want to take another look at the textual descriptions in the template to see if they can better accommodate branch stats when available.

ewjoachim (commented):
Seems awesome so far :)

ferdnyc:
@ewjoachim Could you please fire off another run of the end-to-end tests? I want to get a look at the current comment templates' output in rendered form. Reading it in encoded form in string variables inside the tests is... less-than-enlightening.

ewjoachim:
Wooops totally missed that !

ewjoachim (approved):
/e2e

ferdnyc:
@ewjoachim No worries, I don't know why I asked for it — the current e2e tests don't really show any of the branch coverage features, because they don't include any code changes (just coverage changes).

I was thinking of submitting a (separate) PR that adds an actual _code_ modification to the e2e tests, so that we could get a look at the diff coverage reporting as well. That's the part, if anything, that I think might need adjusting with branch coverage enabled.

(I figured I'd drop the test function to just contain

```python
def f(a="", b="", c="", d=""):
      elements = []
      if a:
          elements.append(a)
      if b:
          elements.append(b)
      """__ADD_CODE_HERE__"""
      return "-".join(elements)
```

...And then have the e2e test function insert the `if c:` and `if d:` blocks. (Or maybe even `if c:` ... `elif d:` which might make for better branch coverage testing.)

WDYT?

ewjoachim:
Good idea :)

ewjoachim:
Hey what would you say that:
- we merge as-is
- I'll be delighted if you want to continue iterate on this

Deal ?

ferdnyc:
@ewjoachim Sure, that works for me. There's nothing that can't be cleaned up in a future PR.

ewjoachim:
Sorry for the delay. I'm all over the place these days. Congratulations for you contribution!

PR #468, opened 2024-09-01
Switch from poetry to uv

ewjoachim:
Of the projects I maintain, this is the one where it's most interesting that installation is as fast as possible, so let's try using uv :) It's an excuse for trying out uv.

Todo if we want to merge:
- Reupload base image
- Update v7 in the different places


ewjoachim:
I'm not particularly surprised, but this doesn't seem to help whatsoever :D

Still might merge, I don't know.

PR #495, opened 2024-10-21
Update release workflow to publish to multiple registries

### LINKED ISSUE #491: Use GitHub Docker Registry
I was wondering if it would be possible to move the Docker container to use [GitHub Docker Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-docker-registry) instead?

I am hitting the Docker Hub pull limit (HTTP 429) in my pipeline a lot which results in failures. GitHub should not have any limits on how much you can pull an image.

ellisvalentiner:
Fixes #491 to publish to both Docker Hub and GitHub Container Registry

Based on documentation:

- https://docs.docker.com/build/ci/github-actions/push-multi-registries/
- https://docs.github.com/en/actions/use-cases-and-examples/publishing-packages/publishing-docker-images

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,6 +15,13 @@ jobs:
     runs-on: ubuntu-latest
     # Check if the tests were successful and were launched by a push event
     if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'push' }}
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
+    env:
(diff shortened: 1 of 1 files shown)


ewjoachim:
Should we also use the newly pushed image ? Or in a subsequent PR ?

ewjoachim (approved):
Seems awesome! Thank you very much!

ewjoachim:
Let's merge as is, it doesn't hurt and it will be much easier to create the part 2 once the part 1 has run.

Issue #529, opened 2025-04-15
Unrelated files may appear in comment


stickperson:
The [diff calculation](https://github.com/py-cov-action/python-coverage-comment-action/blob/ffd1aa12992c2272b950cc6c882674e0c07bc8e8/coverage_comment/coverage.py#L296) is similar to the [two dot diff comparison](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests#three-dot-and-two-dot-git-diff-comparisons) and can result in files unrelated to a PR appearing in the coverage comment. This can happen if there is a merge to the base ref or default branch between the time the GHA was kicked off and this action is run.

I _think_ this is a bug and not the desired result.

ewjoachim:
Oh, you're probably right, it's a bug! Do you want to make a PR ?

stickperson:
Sure, I can get to it today.

stickperson:
This actually may be more complicated than originally planned. I actually think `git fetch origin main --depth=1000` is not doing anything _unless_ `actions/checkout@v4` is using `fetch-depth: 1000` (or 0). This GHA illustrates the point:


```
name: CI

on:
  pull_request:
concurrency:
  group: ${{ github.event_name }}-${{ github.ref }}

jobs:
  fetch_depth:
    name: Fetch all history
    runs-on: ubuntu-latest
    permissions:
      pull-requests: read
      contents: read
    steps:
      - name: Checkout
        uses: actions/checkout@v4
        with:
          fetch-depth: 0
      - name: Fetch base branch
        run: git fetch origin main --depth=1000
      - name: diff
        run: git diff -U0 FETCH_HEAD...
  no_fetch_depth:
    name: Fetch single commit
    runs-on: ubuntu-latest
    permissions:
      pull-requests: read
      contents: read
    steps:
      - name: Checkout
        uses: actions/checkout@v4
      - name: Fetch base branch
        run: git fetch origin main --depth=1000
      - name: diff
        run: git diff -U0 FETCH_HEAD...
```

In `no_fetch_depth`, `git diff -U0 FETCH_HEAD...` fails with `fatal: FETCH_HEAD...HEAD: no merge base`. So I believe the actual fix is:

- remove the [git.fetch](https://github.com/py-cov-action/python-coverage-comment-action/blob/main/coverage_comment/coverage.py#L295) call since it will not do anything
- update the diff command
[...]

PR #532, opened 2025-04-21
Calculate PR-style diff

stickperson:
See #529

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,0 +1,101 @@
+from __future__ import annotations
+
+import os
+import pathlib
+import subprocess
+import uuid
+
+import pytest
+
+
+@pytest.fixture
+def in_integration_env(integration_env, integration_dir):
+    curdir = os.getcwd()
+    os.chdir(integration_dir)
+    yield integration_dir
+    os.chdir(curdir)
+
+
+@pytest.fixture
+def integration_dir(tmp_path: pathlib.Path):
+    test_dir = tmp_path / "integration_test"
+    test_dir.mkdir()
+    return test_dir
+
+
+@pytest.fixture
+def file_path(integration_dir):
+    return integration_dir / "foo.py"
+
+
+@pytest.fixture
+def write_file(file_path):
+    def _(*variables):
+        content = "import os"
+        for i, var in enumerate(variables):
+            content += f"""\nif os.environ.get("{var}"):\n    {i}\n"""
+        file_path.write_text(content, encoding="utf8")
+
+    return _
+
+
+@pytest.fixture
+def run_coverage(file_path, integration_dir):
+    def _(*variables):
+        subprocess.check_call(
+            ["coverage", "run", "--parallel", file_path.name],
+            cwd=integration_dir,
+            env=os.environ | dict.fromkeys(variables, "1"),
+        )
+
+    return _
+
+
+@pytest.fixture
+def commit(integration_dir):
+    def _():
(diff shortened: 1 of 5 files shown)


tests/integration/test_coverage.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
+
+import subprocess
+
+from coverage_comment import coverage
+from coverage_comment import subprocess as subprocess_module
+
+
+def test_get_added_lines(
```
stickperson, line 9: I don't believe this is strictly necessary as it's more testing git than `get_added_lines`. I'm happy to take it out.

tests/integration/conftest.py
```diff
@@ -0,0 +1,101 @@
...
+
+import os
+import pathlib
+import subprocess
+import uuid
+
+import pytest
+
+
+@pytest.fixture
```
ewjoachim, line 11: This and below seems to be copied from test_main, right ?

Could you deduplicate it ? It's ok to move it here, but delete the corresponding copied code from test_main

tests/integration/test_coverage.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    git = subprocess_module.Git()
+    relative_file_path = file_path.relative_to(integration_dir)
+
+    def _check_added_lines():
+        added_lines = coverage.get_added_lines(git, "main")
+
+        assert added_lines == {
+            relative_file_path: list(range(7, 13))  # Line numbers start at 1
+        }
```
ewjoachim, line 23: I'd rather add 2 duplicated lines inside the test than an inner function (so as to make tests as easy to read as possible)

Also, I believe it's ok if you make it a single line:
```python
assert coverage.get_added_lines(git, "main") == {
    relative_file_path: list(range(7, 13))  # Line numbers start at 1
}
```
though that's your choice :)

tests/integration/test_coverage.py
ewjoachim: I'm good keeping it :)

ewjoachim (commented):
I'll try to see why the CI is red.

README.md
```diff
@@ -116,6 +116,10 @@ jobs:
       contents: write
     steps:
       - uses: actions/checkout@v4
+        with:
+          # New lines are discovered by running a three-dot notation diff on base_ref...head. They must share a common
+          # ancestor. Setting fetch-depth to 1000 here should ensure that.
+          fetch-depth: 1000
```
stickperson, line 122: I'll look into the e2e tests to test this since I'm not entirely sure how the actions/checkout action handles this. I wonder if you can get away with not setting `fetch_depth` here since a `git fetch` is called later.

ewjoachim:
I've pushed a PR that fixes pre-commit that was broken, and it wasn't your fault.

ewjoachim (approved):
/e2e

ewjoachim (approved):
/e2e

ewjoachim:
Hey there :) Still interested in getting this merged ?

ewjoachim:
No need for apology :) , I was just worried it slipped your mind, but I'm very fine with a "move slow and don't break people" approach :D

tests/integration/conftest.py
stickperson: Good call. Removed from `test_main`

ewjoachim:
It seems this PR is causing issues, it's probably my fault, but I'll revert for now.

PR #533, opened 2025-04-21
feat: use ghcr image instead of docker

zMynxx:
Avoid a common issue `429 too many requests` due to the super-low DockerHub unauthenticated pulls limit.
Use GHCR instead.

PR #540, opened 2025-05-05
feat: use ghcr image instead of docker

ewjoachim:
This is all @zMynxx's work repackaged, see #533
Thank's a lot to them, it's highly appreciated.


ewjoachim:
Screenshot 2025-05-05 at 09 48 07
Image is pushed and is public, this should work now !

PR #594, opened 2025-08-26
feat: Expose coverage data as action outputs

### LINKED ISSUE #600: Is there a way to fail the job when the MINIMUM_ORANGE is not achieved?
First of all, great GHA!

I would like the CI to fail when the coverage is less than `MINIMUM_ORANGE`, so that users are forced to improve coverage to merge their PR. Is that possible?

https://github.com/MishaKav/pytest-coverage-comment?tab=readme-ov-file#output-example implements a mechanism in which the coverage is an action output, which we can use to fail the CI manually. SOmething like

```
- name: Fail if coverage too low
  if: ${{ steps.coverage.outputs.coverage < 80 }}
  run: |
    echo "Coverage is below 80%!"
    exit 1
```

olivier-lacroix:
Fixes #600

This PR proposes to have the action output (some) coverage data, allowing downstream processing:

 - COVERAGE_PERCENTAGE
 - REFERENCE_COVERAGE_PERCENTAGE


ewjoachim:
Excellent start.

The tests that fail are exactly the tests we would expect to fail.

I hope the contribution experience is pleasant :)

tests/integration/test_main.py
```diff
@@ -111,9 +111,11 @@ def checker(payload):
...
     )
 
-    expected_output = "COMMENT_FILE_WRITTEN=true\n"
-
-    assert output_file.read_text() == expected_output
+    expected_output = "COMMENT_FILE_WRITTEN=true\nCOVERAGE_PERCENTAGE=0.7777777777777778\nREFERENCE_COVERAGE_PERCENTAGE=null\n"
+    # The order of keys is not guaranteed, so we'll sort the lines
+    assert sorted(output_file.read_text().splitlines()) == sorted(
+        expected_output.splitlines()
+    )
```
ewjoachim, line 118:
```suggestion
    expected_output = ["COMMENT_FILE_WRITTEN=true", "COVERAGE_PERCENTAGE=0.7777777777777778", "REFERENCE_COVERAGE_PERCENTAGE=null"]
    # The order of keys is not guaranteed, so we'll sort the lines
    assert sorted(output_file.read_text().strip().splitlines()) == expected_output
```

olivier-lacroix:
I've added a paragraph in the README.

I wonder how this coverage enforcement logic should interact with the merge queue: ideally, you do not want any steps that can fail after the action, otherwise, you risk having incorrect reference data. But, it would feel weird to not enforce coverage in the merge queue. Any thought?

Regarding the second part of your comment, more outputs could definitely be added. I am just not sure which ones, and if you want to make plenty available, maybe they should be made available in a JSON directly?

ewjoachim:
> Regarding the second part of your comment, more outputs could definitely be added. I am just not sure which ones

I'm thinking all the numbers present here (on the badges and their alt-text)
Screenshot 2025-08-27 at 22 51 39

> and if you want to make plenty available, maybe they should be made available in a JSON directly?

Do you have a usecase in mind where it would help to have all of them in a json ? Versus individual variables ? versus... both ?

> I wonder how this coverage enforcement logic should interact with the merge queue: ideally, you do not want any steps that can fail after the action, otherwise, you risk having incorrect reference data. But, it would feel weird to not enforce coverage in the merge queue. Any thought?

I think we should enforce only `on: pull_request`.

olivier-lacroix:
@ewjoachim alright. I have added all the data from `CoverageInfo` (from both the reference branch and the PR branch) and `DiffCoverage` as outputs.

output variable names is built from the classes properties, prefixed with `DIFF` for `DiffCoverage`, `REFERENCE` for `CoverageInfo` data from the base branch and `NEW` for `CoverageInfo` data from the PR. Not 100% happy about these prefixes to be honest:

- we could maybe use no prefix for the `CoverageInfo` data from the PR?
- other suggestion?

TODO:

- [ ] tests to update
- [ ] add an output indicating wich activity was run

olivier-lacroix:
Alright. done. I let you have a look @ewjoachim

ewjoachim (approved):
That seems pretty awesome !

/e2e

ewjoachim:
> we could maybe use no prefix for the `CoverageInfo` data from the PR?

As you have it (`NEW`) or without prefix are both OK to me. Neither are really ambiguous. Having a prefix (like `NEW`) will make it easier to search for (having no prefix would yield both `COVERED_LINES` and `REFERENCE_COVERED_LINES`), so... `NEW` ?

`CURRENT` could also work, but it might be ambiguous with `REFERENCE`. And `BRANCH` is very ambiguous.

I'm tempted to merge as-is. Anything to do before I do ?

(feel free to ping me if I forget to merge by September 9th)

olivier-lacroix:
@ewjoachim gentle ping as requested 🫣

ferdnyc:
At the risk of bikeshedding (and especially late-stage bikeshedding, so double apologies for that) I'm not sure why these are `ALL_CAPS_NAMES` like they're environment variables. That's not typical for workflow outputs, in my experience.

This isn't mere bikeshedding (though I won't deny it's partly that). I'm also worried, as a practical matter, that _because_ they appear to be envvars, people will get confused and try to use them as such.

(IOW, I envision an issue being opened like this:)

> ## Metrics can't be accessed
> 
> I tried to use the new data outputs in a workflow, and nothing showed up. Here's my code:
> 
> ```yaml
> # ...
> 
> steps:
>   - name: "Report Coverage"
>     run: echo "PR coverage: $COVERAGE_PERCENTAGE"
> ```
> 

In GitHub's own examples, outputs are typically lowercase names so they're not confused with envvars. Sometimes, they even _set_ envvars from them, like in [Example: Defining outputs for a job](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-defining-outputs-for-a-job):

> ```yaml
> jobs:
>   job1:
>     runs-on: ubuntu-latest
>     # Map a step output to a job output
>     outputs:
>       output1: ${{ steps.step1.outputs.test }}
>       output2: ${{ steps.step2.outputs.test }}
>     steps:
>       - id: step1
>         run: echo "test=hello" >> "$GITHUB_OUTPUT"
>       - id: step2
>         run: echo "test=world" >> "$GITHUB_OUTPUT"
>   job2:
>     runs-on: ubuntu-latest
>     needs: job1
>     steps:
>       - env:
>           OUTPUT1: ${{needs.job1.outputs.output1}}
>           OUTPUT2: ${{needs.job1.outputs.output2}}
>         run: echo "$OUTPUT1 $OUTPUT2"
> ```

ferdnyc:
(The above being said, I also realize that they're already formatted that way in the code, and it's probably just easier to leave them in the same format for the outputs. I can totally accept that.)

ferdnyc:
@olivier-lacroix 

Since you're effectively writing the same output code twice, for the two dataclasses, what do you think of creating a mixin like this, instead, and just making it a parent class of both dataclasses that can generate outputs?

```python
@dataclasses.dataclass
class OutputMixin:
    """Mixin class for formatting data as GitHub outputs."""
    @staticmethod
    def format_key(key: str, prefix: str) -> str:
        out_name = key.lower().replace("_", "-")
        return f"{prefix}-{out_name}"

    @staticmethod
    def format_value(value):
        if isinstance(value, decimal.Decimal):
            return float(value)
        return value

    def as_output(self, prefix: str) -> dict:
        return {
            self.format_key(key, prefix): self.format_value(value)
            for key, value in dataclasses.asdict(self).items()
            if value is not None and not isinstance(value, dict)
        }
```

As long as the mixin is _also_ a dataclass, even with no data members, static analysis seems OK with the typing.

```python
>>> from coverage_comment import coverage
>>> ci = coverage.CoverageInfo(covered_lines=12,
... num_statements=8, percent_covered=decimal.Decimal(86.4), 
... missing_lines=1, excluded_lines=0)
>>> pprint.pp(ci.as_output("new"))
{'new-covered-lines': 12,
 'new-num-statements': 8,
 'new-percent-covered': 86.4,
 'new-missing-lines': 1,
 'new-excluded-lines': 0,
 'new-num-branches': 0,
 'new-num-partial-branches': 0,
 'new-covered-branches': 0,
 'new-missing-branches': 0}
>>> dc = coverage.DiffCoverage(total_num_lines=2,
... total_num_violations=1,
... total_percent_covered=decimal.Decimal(99.9),
... num_changed_lines=1, files={})
>>> pprint.pp(dc.as_output("diff"))
{'diff-total-num-lines': 2,
 'diff-total-num-violations': 1,
 'diff-total-percent-covered': 99.9,
 'diff-num-changed-lines': 1}
```

olivier-lacroix:
@ewjoachim I have made all outputs lowercase. Note this includes pre-existing `COMMENT_FILE_WRITTEN`, now `comment_file_written`

@ferdnyc fair. I have added an output mixin.

ferdnyc:
GitHub code search shows [324 YAML files containing `COMMENT_FILE_WRITTEN` in various projects](https://github.com/search?q=COMMENT_FILE_WRITTEN+language%3AYAML&type=code&l=YAML). I'm all for renaming, but I think if an _existing_, widely-used output is going to be renamed, there probably needs to be a deprecation period where the old name is also still available, to avoid screwing those users over.

ewjoachim (approved):
/e2e

olivier-lacroix:
Thanks @ewjoachim !

PR #639, opened 2026-02-28
Add an option to generate a plain-text markdown table for PR coverage

dvg-p4:


ewjoachim:
Hello,

The report as it stands today almost reaches the character limits for github comments already.
BTW, you can already [override the template](https://github.com/py-cov-action/python-coverage-comment-action?tab=readme-ov-file#overriding-the-template) if you want.

Maybe I'm missing an important point, in which case feel free to raise it. I don't like deying a contribution on which you've worked, independently of its quality, which is why it's best to open an issue for discussing this first.

dvg-p4:
Oh don't waste your time taking a look at this yet, it's pure vibecoded prototype at the moment. I was just trying to test it on itself, which didn't really work (I missed the fact that there was a whole test repo setup...*facepalm*) 

...yeah using the built-in override functionality would probably be simpler here too, d'oh.

